PR #209, opened 2019-11-22
Refactoring the visualization of deprecated things

lxatrhea:
### Prerequisites

### Description
The default way of hiding/filtering out deprecated things did not always work for treeview controls, like the requirements browser.
In the Requirements browser the list of items that is used to populate the treeview control is a (reactive)List of IRowViewModelBase<T>.
This interface doesn't define the IsDeprecated property, so based on that, the TreeView cannot create the necessary filter for this property, even if the class that implements the interface does implement an IsDeprecated property. In that case the IsDeprecated property surely is there, but it is not "connected" to the interface
For this reason the IsHidden property is added to IRowViewModelBase<T>, so the filter can use that property instead of an IsDeprecated property.
The default value of IsHidden is false, but when a RowViewModelBase<T>, or subclass is created, the RowViewModelBase checks if it implements the IDeprecatableThing interface, or if it's Thing class implements the IDeprecatableThing interface.
If it does, it sets up the lionk between IsDeprecated en IsHidden.

The treeview doesn't hide child nodes based on the fact that a parent node is Deprecated by default. 
Newer versions of DevExpress have more features for this, but, for now, some childnodes need to be hidden "manually" (RequirementsSpecificationRowViewModel).

### DIFF
--- a/CDP4Composition/Mvvm/RowViewModelBase.cs
+++ b/CDP4Composition/Mvvm/RowViewModelBase.cs
@@ -85,7 +85,9 @@ namespace CDP4Composition.Mvvm
         /// Backing field for <see cref="ThingStatus"/>
         /// </summary>
         private ThingStatus thingStatus;
-        
+
+        private ObservableAsPropertyHelper<bool> isHidden;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="RowViewModelBase{T}"/> class. 
         /// </summary>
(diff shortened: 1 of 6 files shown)


CDP4Composition/Mvvm/RowViewModelBase.cs
```diff
@@ -85,7 +85,9 @@ public abstract class RowViewModelBase<T> : ViewModelBase<T>, IRowViewModelBase<
         /// Backing field for <see cref="ThingStatus"/>
         /// </summary>
         private ThingStatus thingStatus;
-        
+
+        private ObservableAsPropertyHelper<bool> isHidden;
```
alexatrhea: Docs please

PR #257, opened 2019-12-18
Replacing editable drop down column by read only text

mmilitaru:
### Prerequisites

- [ ] I have written a descriptive pull-request title
- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/pulls) open

### Description
Changing the editable column with a read only one.

### DIFF
--- a/CDP4SiteDirectory/ViewModels/RoleBrowser/ParticipantPermissionRowViewModel.cs
+++ b/CDP4SiteDirectory/ViewModels/RoleBrowser/ParticipantPermissionRowViewModel.cs
@@ -39,8 +39,6 @@ namespace CDP4SiteDirectory.ViewModels
         public ParticipantPermissionRowViewModel(ParticipantPermission permission, ISession session, IViewModelBase<Thing> containerViewModel)
             : base(permission, session, containerViewModel)
         {
-            this.UpdatePermission();
-            this.WhenAnyValue(x => x.AccessRight).Subscribe(_ => this.ExecuteUpdatePermission());
         }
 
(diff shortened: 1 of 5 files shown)


CDP4SiteDirectory/ViewModels/RoleBrowser/ParticipantPermissionRowViewModel.cs
```diff
@@ -65,6 +65,17 @@ public ClassKind ShortName
             }
         }
 
+        /// <summary>
+        /// Gets the Access right to display it in the "AccessType" column of the Role browser.
+        /// </summary>
+        public string AccessType
```
alexatrhea: The usual way that we do it is instead of introducing a property like this, is we bind to the `this.AccessRight` property and implement rather a Converter that does this `ToString()`. Please do this in this case as well.

CDP4SiteDirectory/ViewModels/RoleBrowser/PersonPermissionRowViewModel.cs
```diff
@@ -66,6 +66,17 @@ public ClassKind ShortName
             }
         }
 
+        /// <summary>
+        /// Gets the Access right to display it in the "AccessType" column of the Role browser.
+        /// </summary>
+        public string AccessType
```
alexatrhea: See previous comment on implementation of converter,

mmilitaru:
Requested changes have been pushed.

PR #275, opened 2019-12-30
refactor FilterStringService: - poprosal

samatrhea:
### Prerequisites

### Description
[WIP] refactor FilterStringService:
  - Use MEF to manage instance and inject in panelNavigationService
  - Use ServiceLocator to get access to the IFilterStringService (as singleton)
  - remove dependency on IDeprecatableToggleViewModel

some work to be done (fix some unit tests); but functionality seems to work again, and IMHO a bit less indirection.

### DIFF
--- a/BasicRdl.Tests/BasicRdlRibbonViewModelTestFixture.cs
+++ b/BasicRdl.Tests/BasicRdlRibbonViewModelTestFixture.cs
@@ -14,6 +14,7 @@ namespace BasicRdl.Tests
     using CDP4Common.SiteDirectoryData;
     using CDP4Composition;
     using CDP4Composition.Navigation;
+    using CDP4Composition.Services;
     using CDP4Composition.Services.FavoritesService;
     using CDP4Dal;
     using CDP4Dal.Events;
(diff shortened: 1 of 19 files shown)


alexatrhea:
In principle it is ok, but imagine the following scenario: I make another toggle button in another plugin. Clicking that toggle will flip the state to `true` in that button and through subscription in the service as well. However the original ribbon button would still be in the state of `false`. You now have decoupled the UI control from the state. It is possible to flip the state without flipping the UI control which could lead to funky behavior. Food for thought.

samatrhea:
> 
> 
> In principle it is ok, but imagine the following scenario: I make another toggle button in another plugin. Clicking that toggle will flip the state to `true` in that button and through subscription in the service as well. However the original ribbon button would still be in the state of `false`. You now have decoupled the UI control from the state. It is possible to flip the state without flipping the UI control which could lead to funky behavior. Food for thought.

indeed, when there is more than one toggle button active in the app, a toggle button that is responsible for toggling deprecatable for all views, then the view-state of those buttons will go out of sync on the first click. Here i would argue that I see little sense in having more than one of these buttons that controls all the views.

MartinFijneman:
when can this be merged? Currently IME implementation does not work for show/hide deprecated items

PR #294, opened 2020-01-23
Refactoring display of arraytype parameters

lxatrhea:

### Description
Arraytype (multivalue type) parameters now show nothing in the UI instead of the string representation of the first value in the array.

### DIFF
--- a/EngineeringModel/ViewModels/ElementDefinitionBrowser/ElementDefinitionTreeRows/ParameterOrOverrideBaseRowViewModel.cs
+++ b/EngineeringModel/ViewModels/ElementDefinitionBrowser/ElementDefinitionTreeRows/ParameterOrOverrideBaseRowViewModel.cs
@@ -269,18 +269,12 @@ namespace CDP4EngineeringModel.ViewModels
             this.Switch = valueSet.ValueSwitch;
             this.Formula = this.GetDisplayValueFromValueSet(valueSet.Formula);
             this.Computed = this.GetDisplayValueFromValueSet(valueSet.Computed);
-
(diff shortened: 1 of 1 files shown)


EngineeringModel/ViewModels/ElementDefinitionBrowser/ElementDefinitionTreeRows/ParameterOrOverrideBaseRowViewModel.cs
```diff
@@ -259,28 +259,48 @@ private ParameterValueSetBase GetValueSet()
...
 
             if (this.ContainedRows.Count == 0)
             {
                 this.ScaleShortName = this.Thing.Scale == null ? "-" : this.Thing.Scale.ShortName;
             }
 
             this.Switch = valueSet.ValueSwitch;
-            this.Formula = valueSet.Formula.Any() ? valueSet.Formula.First() : "-";
-            this.Computed = valueSet.Computed.Any() ? valueSet.Computed.First() : "-";
+            this.Formula = this.GetDisplayValueFromValueSet(valueSet.Formula);
```
samatrhea: i don't understand why the logic for `Formual` and `Computed` is not the same as for `Manual` and `Reference`. Can you explain?

lxatrhea: @samatrhea: I don't know either, but as you can see it was already the case, so someone probably did this on purpose.
I did not want to question this, but if you do... Should we use the same logic for all properties and see what happens?

Issue #290, opened 2020-01-21
Add indentation to first cell in option sheets to emulate a tree like structure


samatrhea:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of the CDP4-IME
- [x] I have searched [open](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues) and [closed](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
The option sheets in MS Excel do not have any indentation which makes the data appear to be a flat list. Indentation needs to be added to the data in the first cell to give the appearance that levels exist and a tree-like structure can be visualized. Based on CDP3 implementation 3 spaces should be added to the text of the name cell per level, where the top-node should not be indented

### Steps to Reproduce
NA

### Logs
NA
  
### System Configuration

- CDP4 version:
  - [x] CDP4-IME: 5.0.2

samatrhea:
I manually verified that this is indeed working as expected

PR #327, opened 2020-02-05
#290 Simulate a tree view by adding indentation for children in the t…

mmilitaru:
…ree.

### Prerequisites

### Description
#290 add indentation to the Excel spreadsheet for Option tab to simulate a tree view structure

### DIFF
--- a/CDP4ParameterSheetGenerator/Assemblers/OptionSheet/OptionSheetRowAssembler.cs
+++ b/CDP4ParameterSheetGenerator/Assemblers/OptionSheet/OptionSheetRowAssembler.cs
@@ -106,11 +106,11 @@ namespace CDP4ParameterSheetGenerator.OptionSheet
                 var nestedElementTreeGenerator = new NestedElementTreeGenerator();
                 nestedElements = nestedElementTreeGenerator.Generate(this.option, this.owner, false);
             }
-            
+
             // Order the nestedElements by shortname. The Shortname of an NestedElement 
             // is related to the path of a NestedElement
             nestedElements = nestedElements.OrderBy(ne => ne.ShortName);
-            
+
             foreach (var nestedElement in nestedElements)
             {
                 var nestedElementExcelRow = new NestedElementExcelRow(nestedElement, this.owner);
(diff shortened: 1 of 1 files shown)


CDP4ParameterSheetGenerator/Assemblers/OptionSheet/OptionSheetRowAssembler.cs
```diff
@@ -120,6 +120,29 @@ public void Assemble()
             foreach (var nestedElementExcelRow in nestedElementExcelRows)
             {
                 var deeprows = nestedElementExcelRow.GetContainedRowsDeep();
+
+                foreach (var itm in deeprows)
+                {
+                    var splitlevel = itm.ModelCode.Split('\\')[0].Split('.').Length;
```
alexatrhea: Is there really no other way to determine the level?

CDP4ParameterSheetGenerator/Assemblers/OptionSheet/OptionSheetRowAssembler.cs
```diff
@@ -120,6 +120,29 @@ public void Assemble()
             foreach (var nestedElementExcelRow in nestedElementExcelRows)
             {
                 var deeprows = nestedElementExcelRow.GetContainedRowsDeep();
+
+                foreach (var itm in deeprows)
```
alexatrhea: Do not shorten variable names. If `itm` meand `item` then use that. But in principle you should use something more descriptive than `item` as well.

CDP4ParameterSheetGenerator/Assemblers/OptionSheet/OptionSheetRowAssembler.cs
```diff
@@ -120,6 +120,29 @@ public void Assemble()
...
+                {
+                    var splitlevel = itm.ModelCode.Split('\\')[0].Split('.').Length;
+
+                    if (itm.Categories == null)
```
alexatrhea: The `Categories` of a Thing should play no role in determining its level.

CDP4ParameterSheetGenerator/Assemblers/OptionSheet/OptionSheetRowAssembler.cs
```diff
@@ -120,6 +120,29 @@ public void Assemble()
...
+                    {
+                        splitlevel += 1;
+                    }
+
+                    var spaces = new string(' ', 3 * Math.Abs(splitlevel - 1));
+
+                    switch (itm)
```
alexatrhea: Is this `switch` necessary? The cases seems to be doing the same thing`

CDP4ParameterSheetGenerator/Assemblers/OptionSheet/OptionSheetRowAssembler.cs
mmilitaru: There is a level property but always comes as zero. Other than that we do not have any thing that can be used directly.

CDP4ParameterSheetGenerator/Assemblers/OptionSheet/OptionSheetRowAssembler.cs
mmilitaru: The categories is used in correlation with the node key in ModelCode. When ther are no categories the ModeCode holds the key in the first segment same as the parent. Otherwise the key is the actual key for instance Loft.Satellite.10 layers while in case of diameter the key is still Loft.Satellite.10 layers and in the second segment stores the name as Loft.Satelitte.10 layers\d\default while "10 layers"is Loft.Satelitte.10 layers\default

CDP4ParameterSheetGenerator/Assemblers/OptionSheet/OptionSheetRowAssembler.cs
mmilitaru: It is since the item is not boxed and the interface has Name property as read only. It cannot be assigned unless we convert to its final type.

alexatrhea:
After discussion, we found another way of doing this. So I close this pull request. Open new PR when that is ready.

Issue #325, opened 2020-02-04
Error while deleting a default option


mmilitaru:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of the CDP4-IME
- [x] I have searched [open](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues) and [closed](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Default option deletion in option iteration browser generates an error.

### Steps to Reproduce
Open Options iterations.
Create a new option and set it as default. Then try to delete that option. Deletion is being completed but an fatal error occurred and there is no more default option when we re-open the app.

### Logs

  - [ ] There are no logs regarding the issue
  - { ] I have attached a log file or included the relevant section in this issue  
  
### System Configuration

- CDP4 version:
  - [x] CDP4-IME: 
  - [ ] Other: 
  -
- Web Services:
  - [ ] CDP4 Web Services version: 
  - [ ] OCDT WSP version: 
- Environment (Operating system, version and so on):
  - [ ] Microsoft Excel (32/64 bit): 
  - [ ] Windows OS (32/64 bit): 
  - [ ] Additional information:

samatrhea:
@mmilitaru Is the error produced by the server? or is it an error that occurs in the IME? Please add logs if you have them

mmilitaru:
Yes, app crashes. And I think this is on the client when retried to setup the default option which is not found.

samatrhea:
@mmilitaru Hi, can you provide log files please so we can go from an "I Think" to an "I know" please?

Issue #107, opened 2019-03-25
Add the shortname of the parametertype to the list of SimpleParameterValues on the Requirements Dialog


samatrhea:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of the CDP4-IME
- [x] I have searched [open](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues) and [closed](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
The Requirements dialog shows the list of SimpleParameterValues. This list should show in the first column the shortname of the referenced ParameterType.

### Steps to Reproduce
NA

### Logs

  - [x] There are no logs regarding the issue
  
### System Configuration

- CDP4 version:
  - [x] CDP4-IME: 2.1.0

alexatrhea:
#206

MartinFijneman:
verified, short name available

PR #395, opened 2020-04-16
Since we accept not to have a default option we must fix it in the ro…

mmilitaru:
…w binding; #325

### Prerequisites

### Description
Fix has applied for a row with no default option

### DIFF
--- a/EngineeringModel/ViewModels/OptionBrowser/OptionBrowserViewModel.cs
+++ b/EngineeringModel/ViewModels/OptionBrowser/OptionBrowserViewModel.cs
@@ -342,11 +342,14 @@ namespace CDP4EngineeringModel.ViewModels
                 return;
             }
 
-            var row = this.Options.Single(x => x.Thing == defaultOption);
-            this.defaultOptionRow = row;
-            row.IsDefaultOption = true;
-        }
+            var row = this.Options.SingleOrDefault(x => x.Thing == defaultOption);
 
+            if (row != null)
+            {
+                this.defaultOptionRow = row;
+                row.IsDefaultOption = true;
+            }
+        }
 
         /// <summary>
         /// Executes the <see cref="SetDefaultCommand"/>


EngineeringModel/ViewModels/OptionBrowser/OptionBrowserViewModel.cs
```diff
@@ -342,11 +342,14 @@ private void UpdateDefaultOption()
                 return;
             }
 
-            var row = this.Options.Single(x => x.Thing == defaultOption);
-            this.defaultOptionRow = row;
-            row.IsDefaultOption = true;
-        }
+            var row = this.Options.SingleOrDefault(x => x.Thing == defaultOption);
```
alexatrhea, line 345: There is a larger problem here. This is not the full solution. If you look at the code you see that if it got to this point it means that `defaultOption` is not null, which means that the iteration still has a `DefaultOption` property set, which, if you deleted the default option alltogether, should not be the case. So to me it looks like the server does not clean up properly when deleting an option. This fix is ok if you have broken data, but in my opinion theres a bug on the server somewhere.

mmilitaru: Yes, the bug on the server side is related to #107 which is already assigned to Kamil. I can take that over and fix it. I know where that is coming from as well.

alexatrhea: issue #107 has nothing to do with this. My guess is the number is wrong?

mmilitaru: Also a bit confused about your comment. The option can be set as default. For instance, we can add multiple options and one is set as default.

mmilitaru: https://github.com/RHEAGROUP/CDP4-WebServices-Community-Edition/issues/107

alexatrhea: After reviewing @kamilatrhea s fix, you are right they are connected.

PR #532, opened 2020-08-06
#507 Parameter finite states

cozminvelciu:
### Prerequisites

### Description
Data source parameter finite states handling:
* one row per parameter per state + one (optional row) for all non-state dependent parameters

### DIFF
--- a/CDP4Reporting.Tests/DataSource/DataSourceParameterTestFixture.cs
+++ b/CDP4Reporting.Tests/DataSource/DataSourceParameterTestFixture.cs
@@ -71,7 +71,7 @@ namespace CDP4Reporting.Tests.DataSource
         [DefinedThingShortName("type1")]
         private class TestParameter1 : ReportingDataSourceParameter<Row>
         {
-            public string GetValue() => this.Value;
+            public string GetValue() => this.ValueSets.First().ActualValue.First();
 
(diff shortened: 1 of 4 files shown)


cozminvelciu:
> 
> 
> I think we should stick with this way for now, but with some additions:
> 
> Add an interface
> ` public interface IReportingDataSourceParameter<T> { T GetValue(); }`
> Let ReportingDataSourceParameter implement this:
> ` public abstract class ReportingDataSourceParameter<T, TType> : ReportingDataSourceColumn<T>, IReportingDataSourceParameter<TType> where T : ReportingDataSourceRow, new()`
> Create a DoubleReportingParameter that derives from `ReportingDataSourceParameter<T, double>` Create a StringReportingParameter that derives from `ReportingDataSourceParameter<T, string>`
> 
> We can use those in reports the way we are used to.

Yes, agreed. This seems like the best approach going forward.

lxatrhea:
OK, let’s doo it like this.
Moving the DefinedShortNameAttribute to property level can then be done at a later stage.

cozminvelciu:
Quite possibly no longer relevant as based on the discussion from 7 August with @lxatrhea the implementation needs to be changed to have the states as columns, now rows.

cozminvelciu:
Closing as obsolete. After the discussions here the parameter finite states were added as columns in #552.

PR #533, opened 2020-08-07
Category skipping levels

cozminvelciu:
### Prerequisites

- [ ] I have written a descriptive pull-request title
- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/pulls) open

### Description
Alternate take on required changes regarding category levels and filtering:
* if the same category is on multiple consecutive levels (possibly with gaps in between), allow it and consider only the leaf-est element
* do not require for categories to strictly match graph depth, and continue traversal until current category is found (regardless of level - e.g. you can find 2nd category in hierarchy on 3rd level)


cozminvelciu:
@lxatrhea went a bit on a tangent and tried to implement the behavior you needed on the Avalon integration branch on top of the previous approach. Does it seem like it fits the bill?

lxatrhea:
> @lxatrhea: I went a bit on a tangent and tried to implement the behavior you needed on the Avalon integration branch on top of the previous approach. Does it seem like it fits the bill?

I should inverstigate further and unfortunately I am not able to take a look at it before my vacation.

cozminvelciu:
A problem I noticed while doing this is that you can have the same element match multiple times from different paths, with no way of differentiating them because now the path is lost due to skipping nodes. (hopefully a clearer example of what I mean is the case of rows[4] and rows[6] in the structure test, which will appear as "identical elements" because the differentiating factor - the 2nd level in the tree - is not displayed at all).

lxatrhea:
> A problem I noticed while doing this is that you can have the same element match multiple times from different paths, with no way of differentiating them because now the path is lost due to skipping nodes. (hopefully a clearer example of what I mean is the case of rows[4] and rows[6] in the structure test, which will appear as "identical elements" because the differentiating factor - the 2nd level in the tree - is not displayed at all).

I think that is indeed the same problem I also had and it was the main reason I stepped to the generator principle.
Again, the problem of building in the constructor is that you first need to create a new node before you can see that you don't need it. This node looks at its children and if one is relevant, then also its parent is relevant.

cozminvelciu:
> 
> 
> > A problem I noticed while doing this is that you can have the same element match multiple times from different paths, with no way of differentiating them because now the path is lost due to skipping nodes. (hopefully a clearer example of what I mean is the case of rows[4] and rows[6] in the structure test, which will appear as "identical elements" because the differentiating factor - the 2nd level in the tree - is not displayed at all).
> 
> I think that is indeed the same problem I also had and it was the main reason I stepped to the generator principle.
> Again, the problem of building in the constructor is that you first need to create a new node before you can see that you don't need it. This node looks at its children and if one is relevant, then also its parent is relevant.

Yes, but from what I understand the same problem will appear with the generator approach as well, as it comes from the way the data is structured (or more exactly allowing levels in the data hierarchy to be skipped), no matter how you parse it.

For the following tree:
```
[     ] +> "top"
["cat"] +-+> "ed1: eu1"
[     ] | +--> "ed2: eu1"
["cat"] | +--> "ed2: eu2"
[     ] +-+> "ed1: eu2"
[     ] | +--> "ed2: eu1"
["cat"] | +--> "ed2: eu2"
```
even with the generator, I think it will still be seen as two "identical" nodes:
* "ed2: eu2" (via "top" -> "ed1: eu1")
* "ed2: eu2" (via "top" -> "ed1: eu2")

CDP4Reporting/DataSource/ReportingDataSourceNode.cs
```diff
@@ -203,6 +208,22 @@ internal static DataTable GetTable(CategoryHierarchy categoryHierarchy)
             }
         }
 
+        /// <summary>
+        /// Checks if the associated <see cref="ElementBase"/> matches the given <see cref="Category"/>.
+        /// TODO check in supercategories also
```
samatrhea, line 213: please make sure when you add a TODO in code there is a corresponding GH ticket so we do not forget to implement it.

Furthermore, what is the intention of matching? I provide some context which may let you decide to rename the method.

an ICategorizable thing can be made a member of a Category, Categories have subcategories and supercategories. When asserting whether an ICategorizable  is a member of a category the super/sub classing in the category model needs to be taken into account. There are methods in the SDK that do this. My impression is that in the code below we are only doing the assert for a specific category, not taking the complete category model into account.

samatrhea (changes requested):

CDP4Reporting/DataSource/ReportingDataSourceNode.cs
lxatrhea: @samatrhea This branche is not going to be merged, because of the architectural difference compared to another branch.
In that branch the SDK is indeed used for category membership checking.

lxatrhea:
Solution is different now. PR can be closed.

PR #588, opened 2020-10-16
DevExpress upgrade and finetuning ReportingParameter default values

lxatrhea:
### Prerequisites

### Description
DevExpress 20.1.8 upgrade and finetuning ReportingParameter default values.


samatrhea (approved):
looks good to me, can you also update the license file please, think we have to make sure that this is in sync with the current state. I mean the AGPL exceptions.

lxatrhea:
> 
> 
> looks good to me, can you also update the license file please, think we have to make sure that this is in sync with the current state. I mean the AGPL exceptions.
The license file only shows the major and minor version numbers. Not the patch, which is the only thing that changed.
So the license file doesn't need changes.

CDP4Reporting/Parameters/ReportingParameter.cs
```diff
@@ -112,7 +118,32 @@ public ReportingParameter(string name, Type type, object defaultValue, string fi
...
+
+        /// <summary>
+        /// Get the default value of a specific <see cref="Type"/>
+        /// </summary>
+        /// <param name="type">The <see cref="Type"/></param>
+        /// <returns>The default value</returns>
+        private object GetDefault(Type type)
+        {
+            return this.GetType()
+                .GetMethod("GetDefaultGeneric", BindingFlags.NonPublic | BindingFlags.Instance)?
```
cozminvelciu, line 135: * it would be better to use `nameof(...)` to get the method name instead of hardcoding it
* the "?" for the null-conditional operator should not be separated from the "." on the next line (it should be on the next line, e.g. `?.MakeGeneric...`)

CDP4Reporting/Parameters/ReportingParameter.cs
```diff
@@ -112,7 +118,32 @@ public ReportingParameter(string name, Type type, object defaultValue, string fi
...
         }
+
+        /// <summary>
+        /// Get the default value of a specific <see cref="Type"/>
+        /// </summary>
+        /// <param name="type">The <see cref="Type"/></param>
+        /// <returns>The default value</returns>
+        private object GetDefault(Type type)
+        {
+            return this.GetType()
```
cozminvelciu, line 134: Wouldn't it be better to get rid of the helper method `GetDefaultGeeneric<T>` and just use as a body for this method (I would even inline this, since it only has one usage):
```
return type.IsValueType ? Activator.CreateInstance(type) : null;
```

cozminvelciu (changes requested):

CDP4Reporting/Parameters/ReportingParameter.cs
lxatrhea: Then I would have a problem with the `bool` type:
A bool is not a value type, so it would default to null using the IsValueType/Activator.CreateInstance method.
default(bool) is false, so there is a big difference.  
I need the check to return `false`, so I'll leave it this way.

cozminvelciu (approved):

alexatrhea (approved):

CDP4Reporting/ViewModels/ReportDesignerViewModel.cs
```diff
@@ -403,10 +403,25 @@ private void CreateNewReport()
...
             this.CurrentReportProjectFilePath = string.Empty;
         }
 
+        /// <summary>
+        /// Get a new <see cref="XtraReport"/> and initialize default settings
+        /// </summary>
+        /// <returns>The <see cref="XtraReport"/></returns>
+        private static XtraReport GetNewXtraReport()
+        {
+            var newReport = new XtraReport();
```
adrianchivu, line 416: Here object initialization might be simplified, but its not important.

adrianchivu (approved):

PR #683, opened 2021-01-13
#631 Support option dependent parameters

cozminvelciu:
Add Crossview support for option dependent parameters.


CDP4CrossViewEditor/Assemblers/CrossviewArrayAssembler.cs
```diff
@@ -75,6 +86,11 @@ public sealed class CrossviewArrayAssembler
         /// </summary>
         public object[,] FormatArray { get; private set; }
 
+        /// <summary>
+        /// Gets the array that contains the parameter sheet information
+        /// </summary>
+        private object[,] FormulaArray { get; set; }
```
cozminvelciu, line 92: This needs to be removed.

Issue #665, opened 2021-01-04
Implement support for compound parameters


cozminvelciu:
The Excel sheet should properly display values for compound parameters.

* [x] implement support for compound parameters
* [ ] update unit tests

cozminvelciu:
See #633 for detailed description.

PR #687, opened 2021-01-15
#665 Support compound parameters

cozminvelciu:
Add Crossview support for single layer compound parameters.

Nested compound parameter types need additional work, but CDP4 doesn't let you create them even though according to 10-25 they are valid.


lxatrhea (approved):

adrianchivu (approved):
Add some test data for increasing coverage.

PR #705, opened 2021-01-19
Fixes #691 and #696 parameter to state mapping

### LINKED ISSUE #691: Parameter to State mapping plugin UI part
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of the CDP4-IME
- [x] I have searched [open](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues) and [closed](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
A user needs to be able to map parameters of an ElementDefinition that have implicit state information described in the name to a state dependent parameter. The use case is the following:

An ElementDefinition (an equipment) stored in a catalogue has the following parameters:
  - power_stby: 5W
  - power_on: 10W
  - power_peak: 20W
  - power_off: 0W

More ElementDefinitions exist with these kind of parameters, all describing according to the data sheets what the power consumption is in the different ElementDefinition specific states (again, described in the naming of the parameters) 
These parameters and the values need to be copied to a study model. 

The target model is a study model where an ActualFiniteStateList is defined which has the following states:
  - Launch mode
  - Observation mode 1
  - Observation mode 2
  - Observation mode 3
  - Safe mode

The requested feature needs to allow a user to determine what the power consumption of a component is, in a specific system mode, reusing the equipment specific power parameters and their values. This mapping will be done by the system-engineer or by the power-engineer, not necessarily by the domain expert that owns the equipment.

The proposed approach is as follows:
  - an equipment will have a **power_consumption** parameter which needs to be state dependent in the study model. This parameter (the ParameterValueSets per state) will contain at the end of the mapping exercise the values of power consumption per state based on the equipment specific power parameters. 
  - an equipment will have a **power_mapping** parameter which is of a specific TextParameterType which will contain the created mapping as its manual value. This mapping will take the form of most likely a JSON document which is the result of a mapping for that equipment
  - the ParamterTypes that describe the different kind of power consumption that may be necessary by the equipments listed in the catalogues need to be added to the Generic Site Reference Data library and need to be categorized with a new Category, something like: "power consumption ParamterTypes".
    - The mapping is performed in a dedicated UI (dialog or docking panel)

The mapping goes along the following procedure:
  1. Select the category of the ElementDefinitions for which the mapping needs to be made (in this case _equipments_) from a combobox
  1. Select the ActualFiniteStateList from a combobox for which the mapping is made from a combox
  1. Select the category of the parameters that need to be mapped (in this case the _power consumption paramtertypes_ category) from a combobox
  1. Select the mapping ParameterType to be used to store the mapping, this can only be a TextParameterType  from a combox
  1. The UI shows a listing of all the ElementUsages (EU) according to the provided filter (selected category) and a table where the mapping can be made. The columns represent the actual states from the selected ActualFiniteStateList, there are 2 rows below each EU. the first row (the mapping row) allows the user to select from a comobox the equipment specific parameter to use for the associated state, the second row (the value row) shows that value of the selected parameter, meaning, the ParameterOverride for that EU.
  1. The user performs the mapping. The mapping is stored in memomory and needs to be updated per EU (using a **update** button per row, a **reset** button is also present that allows a user to go back to a previous mapping in case it exists, if it does not exist then clears the whole selection). When the user hits the **update** button, the mapping is stored in the **power_mapping** parameter and the ValueSets of the **power_consumption** are updated as well

Some extra notes:
  - the ui needs to support repeating the mapping, meaning that the mapping information stored in the TextParameterType needs to include all the required information to be able to do that. Most likely all the info that the user provided in the combobox selection when setting up the mapping.
  - when the values of the equipment specific parameters are updated then the ParameterValueSets of the state-full parameter need to be updated as well, reusing the same mapping. This should be visually apparent by color coding the cells in the second row, meaning the _value row_. By making it orange it should become clear that the values are no longer valid based on the available information.
   - The value of the TextParameter will be written to the Computed Value.

lxatrhea:
### Prerequisites

### Description
Parameter to State Mapping


CDP4ReferenceDataMapper.Tests/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="MSTest.TestAdapter" version="2.1.1" targetFramework="net452" />
```
samatrhea, line 3: i am surprised this file is here, should these references not be in the .csproj file?

CDP4ReferenceDataMapper/Behaviors/MappingCellReadOnlyBehavior.cs
```diff
@@ -0,0 +1,98 @@
...
+    using System.Windows;
+
+    using CDP4ReferenceDataMapper.Managers;
+
+    using DevExpress.Mvvm.UI.Interactivity;
+    using DevExpress.Xpf.Grid;
+    using DevExpress.Xpf.Grid.TreeList;
+
+    /// <summary>
+    /// Add custom readonly behavior to cells of a <see cref="TreeListView"/>
```
samatrhea, line 38: can you correct the docs, not sure how to read the sentence

CDP4ReferenceDataMapper/Converters/MappingRowItemsSourceConverter.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// Not supported
+        /// </summary>
+        /// <param name="value">The parameter is not used.</param>
+        /// <param name="targetTypes">The parameter is not used.</param>
+        /// <param name="parameter">The parameter is not used.</param>
+        /// <param name="culture">The parameter is not used.</param>
+        /// <returns>a <see cref="NotSupportedException"/> is thrown</returns>
+        public object[] ConvertBack(object value, Type[] targetTypes, object parameter, CultureInfo culture)
+        {
+            throw new NotImplementedException();
```
samatrhea, line 82: please make this a `NotSupportedException`

samatrhea (approved):
some minor remarks, would be good to get a demo today

alexatrhea (changes requested):
Some minor things. But can you also straight away include the new plugin into the installer?

alexatrhea (approved):

alexatrhea (approved):

CDP4Composition/ViewModels/ParameterRowControlViewModel.cs
```diff
@@ -199,11 +205,11 @@ protected ParameterValueSetBase GetValueSet()
 
             if (this.Parameter is ParameterOverride parameterOverride)
             {
-                valueSet = parameterOverride.ValueSet.SingleOrDefault(x => (!this.Parameter.IsOptionDependent || (x.ActualOption == this.ActualOption)));
+                valueSet = parameterOverride.ValueSet.FirstOrDefault(x => (!this.Parameter.IsOptionDependent || (x.ActualOption == this.ActualOption)));
```
samatrhea, line 208: if a server is returning incomplete data, i think we fail/crash here. Perhaps wrap an exception or go back to SingleOrDefault?

CDP4CrossViewEditor/Assemblers/CrossviewArrayAssembler.cs
```diff
@@ -93,6 +93,11 @@ int>>>>>> headerDictionary
         /// </summary>
         public object[,] LockArray { get; private set; }
```
samatrhea, line 94: changes to cross-view in this PR?

samatrhea (changes requested):

CDP4CrossViewEditor/Assemblers/CrossviewArrayAssembler.cs
lxatrhea: no, rebased

Issue #712, opened 2021-01-21
Add Crossview dialog checkbox for persisting manually entered values in the sheet


cozminvelciu:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of the CDP4-IME
- [x] I have searched [open](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues) and [closed](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Currently if the user manually changes a value in the Crossview sheet it will be persisted forever in the metadata. The metadata for manually changed values should be cleared at some point.

The Crossview sheet generation dialog should have a checkbox with the label "Persist manually entered values in the sheet" (default on). When this is un-checked, the sheet should disregard values save in the metadata, and clear them.

cozminvelciu:
Having the checkbox unchecked seems to also reset the ElementDefinition/ParameterType saved selection - this should not happen.

cozminvelciu:
Issues reported separately in #727.

PR #684, opened 2021-01-13
#657 Save manually filled values on rebuild

adrianchivu:
### Prerequisites

- [X] I have written a descriptive pull-request title
- [X] I have verified that there are no overlapping [pull-requests](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/pulls) open
- [x] I have verified that I am following the CDP4-IME [code style guidelines](https://raw.githubusercontent.com/RHEAGROUP/CDP4-IME-Community-Edition/master/.github/CONTRIBUTING.md)
- [x] I have provided test coverage for my change (where applicable)

PR #718, opened 2021-01-26
#712 Add crossview dialog checkbox for persisting manually

adrianchivu:
### Prerequisites

- [X] I have written a descriptive pull-request title
- [X] I have verified that there are no overlapping [pull-requests](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/pulls) open


lxatrhea (approved):
I think we should get used to the fact that we should update headers to 2021.

cozminvelciu (changes requested):
Will review after [parent PR](#684) is merged, as for now the diff is inconclusive. Being forced into sequential reviews and merges (instead of handling each issue separately) is one of the disadvantages of having multiple branches based on one another.

CDP4CrossViewEditor/ViewModels/WorkbookSelectionDialogResult.cs
```diff
@@ -90,5 +95,10 @@ public class WorkbookSelectionDialogResult : BaseDialogResult
         /// Get or sets workbook manually edited data values
         /// </summary>
         public Dictionary<string, string> WorkbookChangedValues { get; private set; }
+
+        /// <summary>
+        ///
```
cozminvelciu, line 100: Missing XML docs.

CDP4CrossViewEditor/Views/CrossViewDialog.xaml
```diff
@@ -21,12 +21,19 @@
...
                 <RowDefinition Height="Auto" />
                 <RowDefinition Height="Auto" />
+                <RowDefinition Height="Auto" />
             </Grid.RowDefinitions>
             <Grid.ColumnDefinitions>
                 <ColumnDefinition Width="*" />
             </Grid.ColumnDefinitions>
-            <uc:ElementThingSelector Grid.Row="0" Mode="ElementBase" VerticalAlignment="Stretch" x:Name="ElementThingSelector" />
-            <uc:ParameterThingSelector Grid.Row="1" Mode="ParameterBase" VerticalAlignment="Stretch" x:Name="ParameterThingSelector" />
+            <dxlc:LayoutControl View="Group"
```
cozminvelciu, line 29: One thing I suggest we do is move the checkbox and the text to the bottom of the dialog, and align it to the right, so it jumps out more when clicking OK / Cancel. Where it is now (upper left) makes it very easy to ignore/disregard.

cozminvelciu (changes requested):

CDP4CrossViewEditor/Views/CrossViewDialog.xaml
cozminvelciu: I'm not saying this is the right way, just a suggestion that might be worth implementing _if_ you agree.

cozminvelciu (approved):

Issue #709, opened 2021-01-21
Add proper error handling for sheet generations


cozminvelciu:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of the CDP4-IME
- [x] I have searched [open](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues) and [closed](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
The Crossview sheet generator, along with the Parameter sheet generator, ignore all errors during generation, only logging a small generic error message in the status bar - which is immediately replaced by the end of generation message.

This makes the sheets unusable for:
* the user, as there is no way to know if or what error happens in these modules
* the developers, as there is no way to identify the proper cause of error reports of undesired behavior

A proper error report along with a stacktrace should be displayed.

cozminvelciu:
Note: the error reporting was added only for the Crossview sheet (not for the Parameter sheet also).

PR #777, opened 2021-03-15
#709 Add error handling for the addin

adrianchivu:
### Prerequisites

- [X] I have written a descriptive pull-request title
- [X] I have verified that there are no overlapping [pull-requests](https://github.com/RHEAGROUP/CDP4-IME-Community-Edition/pulls) open
- [X] I have verified that I am following the CDP4-IME [code style guidelines](https://raw.githubusercontent.com/RHEAGROUP/CDP4-IME-Community-Edition/master/.github/CONTRIBUTING.md)
- [X] I have provided test coverage for my change (where applicable)

### DIFF
--- a/CDP4Addin/Addin.cs
+++ b/CDP4Addin/Addin.cs
@@ -35,9 +35,9 @@ namespace CDP4AddinCE
     using System.Reactive.Linq;
     using System.Reflection;
     using System.Runtime.InteropServices;
+    using System.Threading;
     using System.Threading.Tasks;
     using System.Windows;
-    using System.Windows.Threading;
 
     using CDP4AddinCE.Settings;
     using CDP4AddinCE.Utils;
@@ -187,7 +187,8 @@ namespace CDP4AddinCE
             }
             catch (Exception ex)
             {
-                logger.Log(LogLevel.Error, ex);
+                // NOTE: manually handle exceptions as the UI specific dispatcher thread does not work
+                HandleException(ex);
             }
 
             if (this.RibbonUI != null)
@@ -614,8 +615,6 @@ namespace CDP4AddinCE
                 logger.Trace("set theme");
                 ThemeManager.ApplicationThemeName = Theme.SevenName;
                 AppliedTheme.ThemeName = Theme.SevenName;
-
-                Dispatcher.CurrentDispatcher.UnhandledException += UnhandledException;
             }
             catch (CompositionException compositionException)
             {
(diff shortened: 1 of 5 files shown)


cozminvelciu:
The exception report window doesn't act as a modal window; it is started in the background, the focus is not on it, and the excel application keeps running (and is interactable) until it is closed.

alexatrhea (approved):

alexatrhea:
> The exception report window doesn't act as a modal window; it is started in the background, the focus is not on it, and the excel application keeps running (and is interactable) until it is closed.

is this comment regarding behavior as it is implemented in this PR? or as it was before?

adrianchivu:
> > The exception report window doesn't act as a modal window; it is started in the background, the focus is not on it, and the excel application keeps running (and is interactable) until it is closed.
> 
> is this comment regarding behavior as it is implemented in this PR? or as it was before?

It is regarding the behavior. It was implemented in this way because the way how it's implemented in the CDP4IME/Concorde don't work as expected in this case here.

cozminvelciu:
> > The exception report window doesn't act as a modal window; it is started in the background, the focus is not on it, and the excel application keeps running (and is interactable) until it is closed.
> 
> is this comment regarding behavior as it is implemented in this PR? or as it was before?

If I understand your question correctly, it is regarding behavior as it is implemented in this PR, as there was no exception report handling before.

cozminvelciu:
Exceptions are manually handled as the UI specific dispatcher thread method does not work:
```
Dispatcher.CurrentDispatcher.UnhandledException += this.UnhandledExceptionHandler;
```

CDP4Addin/Addin.cs
```diff
@@ -624,6 +632,28 @@ private void AddinOnConnection(object application, ext_ConnectMode connectMode,
...
+        private static void HandleException(Exception ex)
+        {
+            logger.Error(ex);
+
+            var thread = new Thread(() =>
+            {
+                var exceptionReporter = new ExceptionReporter();
+                exceptionReporter.Show(ex);
+
+                Environment.Exit(1);
```
lxatrhea, line 650: Isn't this a big change? Until now every error was "swallowed" and execution continued. Environment.Exit(1) closes the plugin/Excel right? Does it allow the user to save changes to open Excel documents before closing?

adrianchivu: Correct the plugin will be closed. But this action is triggered by the user after he focus the error report window(which doesn't act as a modal window) and click close.
Also the user is able to see and save the error.
And because the error window is started in the background the excel application keeps running and is intractable, until is closed.

cozminvelciu: Yes, until now every error was "swallowed" and execution continued (potentially leaving the process in an undefined state).

`Environment.Exit(1)` does indeed close the Excel process.

Due to the fact that the ExceptionReporter is run on a different thread, the user can still interact with Excel until the exception reporter is closed. To me the fact that the exception reporter lets you still interact with Excel did non seem like a desired behavior, but we couldn't get it to run any other way; I guess we could consider this a feature...

alexatrhea: @cozminvelciu @adrianchivu please remove the `Environment.Exit(1)` for now. Crashing Excel for now is not necessary, at least this was necessary before this change. What is important for us is the error message to analyse.

alexatrhea (changes requested):
Remove the Exit.

lxatrhea (approved):

Issue #960, opened 2021-10-11
Display parameter type categories for parameters in element definition view and product tree


SimonWoodRhea:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of the COMET-IME
- [x] I have searched [open](https://github.com/RHEAGROUP/COMET-IME-Community-Edition/issues) and [closed](https://github.com/RHEAGROUP/COMET-IME-Community-Edition/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Product tree: React to Category changes of ParameterTypes in Parameter Rows.

### Logs

  - [ ] There are no logs regarding the issue
  - { ] I have attached a log file or included the relevant section in this issue  
  
### System Configuration

- COMET version:
  - [ ] COMET-IME: 
  - [ ] Other: 
  -
- Web Services:
  - [ ] CDP4/COMET Web Services version: 
  - [ ] OCDT WSP version: 
- Environment (Operating system, version and so on):
  - [ ] Microsoft Excel (32/64 bit): 
  - [ ] Windows OS (32/64 bit): 
  - [ ] Additional information:

PR #963, opened 2021-10-13
Show parameter type categories in parameter rows feat: #960

SimonWoodRhea:
### Prerequisites

### Description

Display parameter type categories in parameter rows

### DIFF
--- a/EngineeringModel/ViewModels/ElementDefinitionBrowser/ElementDefinitionTreeRows/ElementDefinitionRowViewModel.cs
+++ b/EngineeringModel/ViewModels/ElementDefinitionBrowser/ElementDefinitionTreeRows/ElementDefinitionRowViewModel.cs
@@ -11,14 +11,19 @@ namespace CDP4EngineeringModel.ViewModels
     using System.Linq;
     using System.Threading.Tasks;
     using System.Windows;
+
     using CDP4Common.CommonData;
     using CDP4Common.EngineeringModelData;
     using CDP4Common.SiteDirectoryData;
+
+    using CDP4Composition.Builders;
     using CDP4Composition.DragDrop;
     using CDP4Composition.Mvvm;
     using CDP4Composition.Services;
+
     using CDP4Dal;
     using CDP4Dal.Events;
+
     using CDP4EngineeringModel.Services;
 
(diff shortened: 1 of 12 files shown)


lxatrhea:
> 
> 
> > ElementCategories
> 
> In case of this does anything need to be added to xaml to make filtering work? I think this is better approach. `this.Category` of the param row VM should be left to the PT category, not the categories of the parent. And indeed we let the relevant parties know about the changes.

Well, the Filtering works based on a single FieldName that should be the same for all rows:
```
                        <dxg:TreeListColumn FieldName="Category" Header="Category" Visible ="False" ShowInColumnChooser="False">
                            <dxmvvm:Interaction.Behaviors>
                                <behaviors:FilterOperatorBehavior CustomFilterOperatorType="Category" ItemsSource="{Binding ElementDefinitionRowViewModels}"/>
                            </dxmvvm:Interaction.Behaviors>
                        </dxg:TreeListColumn>
```
I would propose to renaming the Category property used for filtering on all possible RowViewModel types in the ProductTree and ElementDefinition Browser that represent a Categorizable Thing to `ElementCategory`. 

Next to that we implement a Category property that implements the Thing's Category property 1-on-1. This one is used in the UI.

SimonWoodRhea:
@lxatrhea @alexatrhea 
What would that mean for Parameter rows? Since they don't represent a categorizable thing, which category property should be used? 

If ElementCategory is used it will be filterable but not displayed.
If Category is used it will be displayed but not filterable.

So it seems both fields should be used.

EngineeringModel/ViewModels/ElementDefinitionBrowser/ElementDefinitionTreeRows/ElementDefinitionRowViewModel.cs
```diff
@@ -74,6 +79,24 @@ public IThingCreator ThingCreator
             set => this.thingCreator = value;
         }
 
+        /// <summary>
+        /// Gets the Categories in display format
+        /// </summary>
+        public string DisplayCategory
```
lxatrhea, line 85: This property isn't Reactive, but I think it should be.

PR #981, opened 2021-11-01
[IMPROVE] Initialize BrowserViewModels using Dispatcher; Fixes #978

### LINKED ISSUE #978: Huge delay in UI response for Element Definitions tab under Model view
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of the COMET-IME
- [x] I have searched [open](https://github.com/RHEAGROUP/COMET-IME-Community-Edition/issues) and [closed](https://github.com/RHEAGROUP/COMET-IME-Community-Edition/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
There is an unacceptable delay between pressing the "Element Definitions" icon in the Model view ribbon and opening of the tab that contains the info. In our set-up this takes 3 to 10 seconds, with NO response in the UI. The problem is that the user *thinks* that he didn't press the button, as there is no visual feedback; this leads to behavior where users will press the button multiple times, leading to even longer delays in response. Note: the "busy loading" indicator only becomes visible AFTER the Element Definitions tab is shown in the UI.

It seems that the loading takes place in the UI thread and hence blocks the update of the UI itself.

### Steps to Reproduce

1. Open any large model
2. Go to Model View
3. Press "Element Definitions" icon in the ribbon
4. The Element Definitions tab appears after 3-10 seconds, depending on size of the model.

### Logs

No logs provided. Tested both on CITRIX as well as "native" configurations - almost same behavior. Problem is not related to local client setup - it is reproducible on multiple machines. Likewise the issue is not server related, we have tested it on two servers, and get same behavior. Server performance is nominal, as are data loading times reported back in the UI.
  
### System Configuration

- COMET version:
  - [x] COMET-IME: v9.0.1
- Web Services:
  - [x] CDP4/COMET Web Services version: v7.0.2
- Environment (Operating system, version and so on):
  - [x] Windows OS (32/64 bit): W10 (fully up-to-date)
  - [x] Additional information: connected to server over VPN (but seems irrelevant - other features work fine, and data is already loaded, so it is a pure IME local thing)

lxatrhea:
### Prerequisites

### Description
Show loading indication for Element Definitions browser, Requirements Browser and Product Tree.

### DIFF
--- a/CDP4Composition/Mvvm/BrowserViewModelBase.cs
+++ b/CDP4Composition/Mvvm/BrowserViewModelBase.cs
@@ -163,6 +163,11 @@ namespace CDP4Composition.Mvvm
         private bool isFilterEnabled;
 
         /// <summary>
+        /// Backing field for <see cref="LoadingMessage"/>
+        /// </summary>
+        private string loadingMessage = string.Empty;
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="BrowserViewModelBase{T}"/> class.
         /// Used by MEF.
         /// </summary>
@@ -334,6 +339,15 @@ namespace CDP4Composition.Mvvm
         {
             get { return this.canWriteSelectedThing; }
             set { this.RaiseAndSetIfChanged(ref this.canWriteSelectedThing, value); }
+        }
+
+        /// <summary>
+        /// Gets or sets the loading message
+        /// </summary>
+        public string LoadingMessage
+        {
+            get => this.loadingMessage;
+            protected set => this.RaiseAndSetIfChanged(ref this.loadingMessage, value);
         }
 
         public Dictionary<DataViewBase, Dictionary<string, (CustomFilterOperatorType, IEnumerable<IRowViewModelBase<Thing>>)>> CustomFilterOperators { get; } = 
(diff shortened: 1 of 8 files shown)


alexatrhea (approved):
Curious to see this working

CDP4Composition/Mvvm/BrowserViewModelBase.cs
```diff
@@ -1044,5 +1058,47 @@ private void PopulateCreateContextMenu()
             this.CreateContextMenu.AddRange(this.ContextMenu.Where(x => x.MenuItemKind == MenuItemKind.Create));
             this.IsAddButtonEnabled = this.CreateContextMenu.Any();
         }
+
+        /// <summary>
+        /// Executes a long running <see cref="Action"/> using the current <see cref="Dispatcher"/> async.
+        /// </summary>
+        /// <param name="action">The <see cref="Action"/></param>
+        /// <param name="loadingMessage">A message to be shown during execution.</param>
+        protected void ExecuteLongRunningDispatcherAction(Action action, string loadingMessage = "")
```
samatrhea, line 1067: should this not return a Task instead of void?

Requirements/ViewModels/RequirementBrowser/RequirementsBrowserViewModel.cs
```diff
@@ -153,21 +153,26 @@ public class RequirementsBrowserViewModel : ModellingThingBrowserViewModelBase,
...
 
-            this.AddSubscriptions();
-            this.UpdateProperties();
+                    this.UpdateRequirementSpecificationsRows();
 
-            this.openRequirementsSpecificationEditorViewModels = new List<RequirementsSpecificationEditorViewModel>();
+                    this.AddSubscriptions();
+                    this.UpdateProperties();
+                },
+                "Loading browser...");
```
samatrhea, line 175: i propose we make this specific to the browser, please rephrase to `Loading Requirements Browser`. I think the `PanelCaption` can be used here.

ProductTree/ViewModels/ProductTreeViewModel.cs
```diff
@@ -150,13 +150,17 @@ public ProductTreeViewModel(Option option, ISession session, IThingDialogNavigat
...
 
-            this.UpdateProperties();
+            this.ExecuteLongRunningDispatcherAction(
+                () => 
+                { 
+                    this.AddSubscriptions();
+                    this.SetTopElement(iteration);
+                    this.UpdateProperties();
+                },
+                "Loading browser...");
```
samatrhea, line 163: i propose we make this specific to the browser, please rephrase to `Loading Product Tree`. I think the `PanelCaption` can be used here

samatrhea (changes requested):
code looks good, some question about returning void or Task on the method that makes loading the viewmodel run in the background. We need to make sure that exceptions are not swallowed and bubble up the way they would if this was not executed on the background.

I tested it both in the IME and in Excel, all working

CDP4Composition/Mvvm/BrowserViewModelBase.cs
lxatrhea: No, because nothing should be awaited. 
The trick is that we do a "fire and forget" invocation of some code and we prioritize the code to be executed in the Dispatcher stack (DispatcherPriority.Background). 
Because of the lower priority, the Dispatcher handles rendering the loading control (IsBusy = true) first. 
When the delayed code starts executing (on the UI thread) the code will block the UI thread again.

If you use the Dispatcher extensively, you're in deep s***, but for actions like this it works perfectly.

samatrhea: ok

samatrhea (approved):

PR #1163, opened 2022-09-13
Feat/gh 309 add warning to person dialog when creating a person with no password

jan-at-rhea:
### Prerequisites

### Description
Warning if password is not set.
image

Also fixed a small issue that was raising exception if the password button was clicked twice.

### DIFF
--- a/CDP4SiteDirectory/ViewModels/Dialogs/PersonDialogViewModel.cs
+++ b/CDP4SiteDirectory/ViewModels/Dialogs/PersonDialogViewModel.cs
@@ -109,7 +109,11 @@ namespace CDP4SiteDirectory.ViewModels
         public bool PwdEditIsChecked
         {
             get { return this.pwdEditIsChecked; }
-            set { this.RaiseAndSetIfChanged(ref this.pwdEditIsChecked, value); }
+            set
+            {
+                this.RaiseAndSetIfChanged(ref this.pwdEditIsChecked, value);
+                this.RaisePropertyChanged(nameof(this.ShoudDisplayPasswordNotSetWarning));
+            }
         }
 
         /// <summary>
@@ -130,7 +134,15 @@ namespace CDP4SiteDirectory.ViewModels
         /// <summary>
         /// Gets the <see cref="ICommand"/> to set the default <see cref="EmailAddress"/>
         /// </summary>
-        public ReactiveCommand<object> SetDefaultEmailAddressCommand { get; private set; } 
+        public ReactiveCommand<object> SetDefaultEmailAddressCommand { get; private set; }
+        
+        /// <summary>
+        /// Returns true if new <see cref="Person"/> nas no passwords set
+        /// </summary>
+        public bool ShoudDisplayPasswordNotSetWarning
+        {
+            get { return this.dialogKind == ThingDialogKind.Create && this.PwdEditIsChecked; }
+        }
 
         /// <summary>
         /// Gets the error message for the property with the given name.
--- a/CDP4SiteDirectory/ViewModels/Dialogs/PersonDialogViewModel.cs
+++ b/CDP4SiteDirectory/ViewModels/Dialogs/PersonDialogViewModel.cs
@@ -146,7 +146,7 @@ namespace CDP4SiteDirectory.ViewModels
                 if ((columnName == "Password" || columnName == "PasswordConfirmation") && !this.PwdEditIsChecked)
                 {
                     var validationErrorToRemove =
-                        this.ValidationErrors.SingleOrDefault(
+                        this.ValidationErrors.FirstOrDefault(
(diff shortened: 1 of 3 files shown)


CDP4SiteDirectory/ViewModels/Dialogs/PersonDialogViewModel.cs
```diff
@@ -146,7 +146,7 @@ public override string ShortName
                 if ((columnName == "Password" || columnName == "PasswordConfirmation") && !this.PwdEditIsChecked)
                 {
                     var validationErrorToRemove =
-                        this.ValidationErrors.SingleOrDefault(
+                        this.ValidationErrors.FirstOrDefault(
```
jan-at-rhea, line 149: Fixes exception when double clicking.

CDP4SiteDirectory/Views/Dialogs/PersonDialog.xaml
```diff
@@ -19,7 +20,7 @@
     <dx:DXWindow.Resources>
         <ResourceDictionary>
             <cdp4Composition:ThingToIconUriConverter x:Key="ThingToIconUriConverter" />
-            <BooleanToVisibilityConverter x:Key="BooleanToVisibilityConverter" />
+            <cdp4Converters:BooleanToVisibilityConverter x:Key="BooleanToVisibilityConverter" />
```
jan-at-rhea, line 23: This used the standard MS converter, not the one stored in cdp4. Had to change it in order to inverse Boolean.

CDP4SiteDirectory/Views/Dialogs/PersonDialog.xaml
```diff
@@ -230,7 +231,16 @@
                         <dxe:PasswordBoxEdit Name="PasswordConfirmation" Text="{Binding Path=PasswordConfirmation, Mode=TwoWay, ValidatesOnDataErrors=True, UpdateSourceTrigger=PropertyChanged}" />
                     </Grid>
                 </lc:LayoutItem>
-
+                <lc:LayoutItem Label=" " Visibility="{Binding PwdEditIsChecked, Converter={StaticResource BooleanToVisibilityConverter }, ConverterParameter=Invert }">
```
alexatrhea, line 234: You can remove the `Label` attribute all together.

Further, I'm not exactly super sure about the visibility logic here. To me this warning should be showing on two conditions being true:

- The Person is being created (when editing an existing Person there is no way clientside to know whether the password is set)
- The password is NOT being set so when this boolean is False (if I understand correctly)

alexatrhea (changes requested):

CDP4SiteDirectory/Views/Dialogs/PersonDialog.xaml
lxatrhea: I agree. The thing we want to achieve is a reminder to the user that the password has not been set at creation time. On update we don't know the password in the UI as existing passwords are not read from the WebService.

lxatrhea (changes requested):

CDP4SiteDirectory/Views/Dialogs/PersonDialog.xaml
jan-at-rhea: @alexatrhea I don't like how it looks without the label parameter:
image
Somehow I prefer it to be aligned with the text box above. What do you think?

alexatrhea: ah in this case I understand. Disregard then.

jan-at-rhea:
@alexatrhea @lxatrhea Changed the binding to new property that is true only if edit mode == create and password is not set

alexatrhea (approved):

CDP4SiteDirectory/ViewModels/Dialogs/PersonDialogViewModel.cs
```diff
@@ -109,7 +109,11 @@ public string PasswordConfirmation
         public bool PwdEditIsChecked
         {
             get { return this.pwdEditIsChecked; }
-            set { this.RaiseAndSetIfChanged(ref this.pwdEditIsChecked, value); }
+            set
+            {
+                this.RaiseAndSetIfChanged(ref this.pwdEditIsChecked, value);
+                this.RaisePropertyChanged(nameof(this.ShoudDisplayPasswordNotSetWarning));
```
lxatrhea, line 115: I think we should implement this using ReactiveUI and having a third property as described here: https://stackoverflow.com/questions/27581449/the-correct-way-to-raise-propertychanged-in-reactiveui

Or just create a WhenAnyValue statement on the ShoudDisplayPasswordNotSetWarning property that checks the DialogKind and PwdEditIsChecked property. Then ShouldDisplayPasswordNotSetWarning will also have a setter.

lxatrhea (changes requested):

PR #1459, opened 2026-07-24
Feat/gh1458 possible to drag and drop usages to re-parent them and bumped devexpress to 23.2.15

### LINKED ISSUE #1458: [Feat] As a User I want to move an element usage from one element definition into another while holding all its values and overrides
As a User I want to move an element usage from one element definition into another while holding all its values and override values, such that it is easier to change the architecture and not lose information. Keep the usage name and shortname, and keep the values and override values.

rdevoogtatstariongroup:
### Prerequisites

## Move an Element Usage to another Element Definition (drag & drop)  closes #1458

Adds drag-and-drop **reparenting** of an `ElementUsage`. Drag a usage onto an
`ElementDefinition` (in the Element Definitions browser or the Product Tree; in the
Product Tree also onto another usage) and it is **moved** into that definition,
keeping its `Iid`, name, short-name, owner, excluded options and all
`ParameterOverride`s (values included). Previously the only workaround was to
re-create the usage manually, losing all of that.

### How
- New `IThingCreator.MoveElementUsage(usage, targetDefinition, session)` re-parents the
  usage in a single `ThingTransaction` (same pattern as
  `RequirementRowViewModel.ChangeRequirementContainer`).
- New shared `CDP4Composition/DragDrop/ElementUsageDropValidator` centralizes the drop
  guards so both browsers stay consistent and cannot create invalid models:
  write permission, same-model only, and `ElementDefinition.HasUsageOf(...)` to block
  self-drop and transitive containment loops.
- Wired into `ElementDefinitionRowViewModel` (EngineeringModel + ProductTree) and the
  ProductTree `ElementUsageRowViewModel`; a tooltip documents the gesture.

### Not included
Duplicating a usage is intentionally left out: the data-source only supports a
server-side copy of an `ElementDefinition` ("Only Element-Definition may be copied"),
and a client-side deep-create could not reliably restore override values. Tracked
separately (see linked web/SDK tickets); **move already preserves everything**.

### Note for reviewers
This branch also bumps **DevExpress 23.2.5 → 23.2.15** across all `.csproj` to fix the
local `NU1605` restore conflict. CI/AppVeyor must serve 23.2.15. Happy to split this
into its own commit/PR if preferred.


joao4all (approved):
LGTM

CDP4Composition/Services/ThingCreator.cs
```diff
@@ -269,7 +269,58 @@ public async Task CreateElementUsage(ElementDefinition container, ElementDefinit
...
+            }
+
+            // re-parent the usage by adding it to the target definition; the same pattern is used to move a Requirement between specifications.
+            var transactionContext = TransactionContextResolver.ResolveContext(targetElementDefinition);
+            var transaction = new ThingTransaction(transactionContext);
+
+            var usageClone = elementUsage.Clone(false);
+            transaction.CreateOrUpdate(usageClone);
+
+            var elementDefinitionClone = targetElementDefinition.Clone(false);
```
samatstariongroup, line 311: just to be sure, do we have to remove the usage from the old container? or does the containerlist already take care of this?

samatstariongroup (changes requested):

CDP4Composition/Services/ThingCreator.cs
rdevoogtatstariongroup: No explicit removal is needed: an ElementUsage has a single container, so adding the clone to the target definition's ContainedElement re-parents it and the data-source moves it out of the old container automatically (same pattern as RequirementsGroupRowViewModel.ChangeRequirementGroup), confirmed working in both the Element Definitions browser and the Product Tree.

rdevoogtatstariongroup: No explicit removal is needed: an ElementUsage has a single container, so adding the clone to the target definition's ContainedElement re-parents it and the data-source moves it out of the old container automatically (same pattern as RequirementsGroupRowViewModel.ChangeRequirementGroup), confirmed working in both the Element Definitions browser and the Product Tree.

PR #1467, opened 2026-08-13
Feat #1466 Configurable relationship traceability diagram

### LINKED ISSUE #1466: [Feature] As a user I want to view relationships in configurable diagram
## Description

Add a **configurable relationship traceability diagram** to the IME: pick a starting point, walk
`Relationship`s a configurable number of levels up and down, and render the result on the
existing diagram surface. Read-only viewer — nothing is written to the model.

Today this is only achievable by scripting the model out to DOT/HTML, which is per-project,
manual and disconnected from the IME.

It must be **generic** (no assumption about a project's architecture or category names) and
**filterable** (a full model is unreadable otherwise). Configuration should mirror the
Relationship Matrix, which already solves "pick things by ClassKind + Category, pick a
relationship rule".

**Configuration**
- **Root**: a selected `Thing`, *or* a set defined by ClassKind + Category(ies).
- **Depth**: levels down and levels up, independently (e.g. 3 down / 1 up).
- **Per-level relationship filter**: a default filter applied to all levels, with optional
  per-level overrides. A filter = a relationship rule and/or relationship `Category`, plus
  direction.
- **Node filter**: optionally only show reached things of a given ClassKind/Category.
- **Presets**: save/load a named configuration (as the Relationship Matrix does).
- **Guard**: stop at a max node count and warn, instead of rendering an unusable hairball.

Both `BinaryRelationship` (directed: Source→Target down, Target→Source up) and
`MultiRelationship` (undirected: entering any `RelatedThing` expands to the others, costing one
level) are traversed.

**Typical uses**: requirement→requirement decomposition, interface traceability,
requirement→element allocation.

**Reuse**
- `RelationshipMatrix`: `SourceConfigurationViewModel`, `RelationshipConfigurationViewModel`,
  plugin-settings preset pattern.
- `CDP4DiagramEditor`: diagram control, node/edge view models, drag-drop, selection wiring.

## Acceptance criteria

- [ ] A graph builder walks `BinaryRelationship` and `MultiRelationship` from a root, honouring
      separate up/down depth limits, per-level filters and a node filter; cycles terminate;
      unit tested.
- [ ] A new panel renders the result with auto-layout; node label = short name, tooltip =
      definition; selecting a node shows the Thing in the property grid.
- [ ] Configuration is editable in the panel and re-renders live; no project-specific
      category/ClassKind is hard-coded.
- [ ] A configuration can be saved and reloaded as a named preset.
- [ ] Exceeding the max node count shows a warning instead of hanging the UI.

rdevoogtatstariongroup:
### Prerequisites

### Description
Fix #1466 Configurable relationship traceability diagram

Adds a new Relationship Traceability panel to the Diagram Editor plugin. It renders a read-only diagram of everything reachable from a set of root things by walking BinaryRelationships and MultiRelationships, without writing anything to the model.

* Roots by drag and drop from any browser and/or by class kind and category selection, mirroring the Relationship Matrix configuration.
* Separate depth limits for downstream and upstream traversal, with a maximum node guard and a warning when it trips.
* Relationship filtering by category, with a configurable arrow direction (along, against, both) as default and per level, so hierarchies whose relationship arrows change direction along the chain can be shown in one diagram.
* Sugiyama auto layout in four orientations; connectors run in traversal level order while arrow heads keep the true relationship direction; undirected multi relationships are drawn without arrow heads.
* Right click to exclude a node together with everything only reachable through it, with a restore list in the panel.
* Selecting a node publishes it to the property grid.
* Named presets saved through the plugin settings service.
* Export to PNG, JPEG and SVG (SVG is generated from the laid out items because the built-in export skips templated node content).

The traversal lives in a UI-free RelationshipGraphBuilder with pre-indexed relationship lookups. The configuration pane uses the same NavBar/LayoutControl arrangement as the Relationship Matrix.

image

### DIFF
--- a/CDP4DiagramEditor/Behaviors/TraceabilityDiagramBehavior.cs
+++ b/CDP4DiagramEditor/Behaviors/TraceabilityDiagramBehavior.cs
@@ -52,10 +52,10 @@ namespace CDP4DiagramEditor.Behaviors
     public class TraceabilityDiagramBehavior : Behavior<DiagramControl>, ITraceabilityDiagramBehavior
     {
         /// <summary>
-        /// A value indicating whether a layout pass is already scheduled, used to coalesce the per-item
-        /// <see cref="DiagramControl.ItemsChanged"/> events into a single layout pass
+        /// The scheduled layout pass, used both to coalesce the per-item <see cref="DiagramControl.ItemsChanged"/>
+        /// events into a single layout pass and to abort that pass when the behavior detaches
         /// </summary>
-        private bool layoutPending;
+        private DispatcherOperation layoutOperation;
 
         /// <summary>
         /// The on attached event handler
@@ -73,6 +73,10 @@ namespace CDP4DiagramEditor.Behaviors
         /// </summary>
         protected override void OnDetaching()
         {
+            // a layout pass queued by the last items change would run after the panel is gone
+            this.layoutOperation?.Abort();
+            this.layoutOperation = null;
+
             this.AssociatedObject.DataContextChanged -= this.OnDataContextChanged;
             this.AssociatedObject.ItemsChanged -= this.OnItemsChanged;
 
@@ -90,7 +94,7 @@ namespace CDP4DiagramEditor.Behaviors
         /// </summary>
         public void ApplyLayout()
         {
-            if (!this.AssociatedObject.Items.OfType<DiagramContentItem>().Any())
+            if (this.AssociatedObject == null || !this.AssociatedObject.Items.OfType<DiagramContentItem>().Any())
             {
                 return;
             }
@@ -237,17 +241,15 @@ namespace CDP4DiagramEditor.Behaviors
                 connector.EndArrow = edge.IsReversed || edge.IsUndirected ? null : ArrowDescriptions.Filled90;
             }
 
-            if (this.layoutPending)
+            if (this.layoutOperation != null)
             {
                 return;
             }
 
-            this.layoutPending = true;
-
-            this.AssociatedObject.Dispatcher.BeginInvoke(
+            this.layoutOperation = this.AssociatedObject.Dispatcher.BeginInvoke(
                 new Action(() =>
                 {
-                    this.layoutPending = false;
+                    this.layoutOperation = null;
                     this.ApplyLayout();
                 }),
                 DispatcherPriority.Background);
--- a/CDP4DiagramEditor/Behaviors/TraceabilityDiagramBehavior.cs
+++ b/CDP4DiagramEditor/Behaviors/TraceabilityDiagramBehavior.cs
@@ -1,0 +1,256 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="TraceabilityDiagramBehavior.cs" company="Starion Group S.A.">
+//    Copyright (c) 2015-2026 Starion Group S.A.
+//
+//    Author: Sam Gerené, Alex Vorobiev, Alexander van Delft, Nathanael Smiechowski, Antoine Théate, Rowan de Voogt
+//
+//    This file is part of CDP4-COMET IME Community Edition.
+//    The CDP4-COMET IME Community Edition is the Starion Concurrent Design Desktop Application and Excel Integration
+//    compliant with ECSS-E-TM-10-25 Annex A and Annex C.
+//
+//    The CDP4-COMET IME Community Edition is free software; you can redistribute it and/or
+//    modify it under the terms of the GNU Affero General Public
+//    License as published by the Free Software Foundation; either
+//    version 3 of the License, or any later version.
+//
+//    The CDP4-COMET IME Community Edition is distributed in the hope that it will be useful,
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+//    GNU Affero General Public License for more details.
+//
+//    You should have received a copy of the GNU Affero General Public License
+//    along with this program. If not, see http://www.gnu.org/licenses/.
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace CDP4DiagramEditor.Behaviors
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+    using System.Security;
+    using System.Text;
+    using System.Windows;
+    using System.Windows.Threading;
+
+    using CDP4Composition.Navigation;
+
+    using CDP4DiagramEditor.ViewModels;
+
+    using CommonServiceLocator;
+
+    using DevExpress.Diagram.Core;
+    using DevExpress.Mvvm.UI.Interactivity;
+    using DevExpress.Xpf.Diagram;
+
+    /// <summary>
+    /// The view side of the traceability diagram: it applies the Sugiyama automatic layout in the orientation selected
+    /// on the <see cref="RelationshipTraceabilityViewModel"/> whenever the diagram items change, and exports the
+    /// diagram to a file
+    /// </summary>
+    public class TraceabilityDiagramBehavior : Behavior<DiagramControl>, ITraceabilityDiagramBehavior
+    {
+        /// <summary>
+        /// A value indicating whether a layout pass is already scheduled, used to coalesce the per-item
+        /// <see cref="DiagramControl.ItemsChanged"/> events into a single layout pass
+        /// </summary>
+        private bool layoutPending;
+
+        /// <summary>
+        /// The on attached event handler
+        /// </summary>
+        protected override void OnAttached()
+        {
+            base.OnAttached();
+            this.AssociatedObject.DataContextChanged += this.OnDataContextChanged;
+            this.AssociatedObject.ItemsChanged += this.OnItemsChanged;
+            this.OnDataContextChanged(this.AssociatedObject, default);
+        }
+
+        /// <summary>
+        /// Unsubscribes event handlers when detaching
+        /// </summary>
+        protected override void OnDetaching()
+        {
+            this.AssociatedObject.DataContextChanged -= this.OnDataContextChanged;
+            this.AssociatedObject.ItemsChanged -= this.OnItemsChanged;
+
+            if (this.AssociatedObject.DataContext is RelationshipTraceabilityViewModel viewModel)
+            {
+                viewModel.Behavior = null;
+            }
+
+            base.OnDetaching();
+        }
+
+        /// <summary>
+        /// Applies the Sugiyama automatic layout in the orientation currently selected on the view-model and centers
+        /// the page
+        /// </summary>
+        public void ApplyLayout()
+        {
+            if (!this.AssociatedObject.Items.OfType<DiagramContentItem>().Any())
+            {
+                return;
+            }
+
+            var direction = (this.AssociatedObject.DataContext as RelationshipTraceabilityViewModel)?.LayoutDirection ?? Direction.Down;
+
+            this.AssociatedObject.ApplySugiyamaLayout(direction, this.AssociatedObject.Items);
+            this.AssociatedObject.AlignPage(HorizontalAlignment.Center, VerticalAlignment.Center);
+        }
+
+        /// <summary>
+        /// Exports the diagram to a file picked by the user
+        /// </summary>
+        /// <param name="format">The <see cref="DiagramExportFormat"/> to export to</param>
+        public void Export(DiagramExportFormat format)
+        {
+            var openSaveFileDialogService = ServiceLocator.Current.GetInstance<IOpenSaveFileDialogService>();
+            var extension = format.ToString().ToLower();
+            var result = openSaveFileDialogService.GetSaveFileDialog($"RelationshipTraceability-{DateTime.Now:yyyy-MM-dd_HH-mm}", $".{extension}", $"{format} file(*.{extension}) | *.{extension}; ", "", 0);
+
+            if (string.IsNullOrWhiteSpace(result))
+            {
+                return;
+            }
+
+            if (format == DiagramExportFormat.SVG)
+            {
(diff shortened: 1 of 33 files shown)


CDP4DiagramEditor/Helpers/RelationshipGraphBuilder.cs
```diff
@@ -0,0 +1,341 @@
...
+
+            var nodes = new Dictionary<Guid, RelationshipGraphNode>();
+            var edges = new List<RelationshipGraphEdge>();
+            var edgeKeys = new HashSet<Tuple<Guid, Guid, Guid>>();
+            var rootThings = new List<Thing>();
+            var maxNodeCountReached = false;
+
+            foreach (var root in roots.Where(x => x != null))
+            {
+                if (nodes.ContainsKey(root.Iid) || configuration.ExcludedThings.Contains(root.Iid))
```
samatstariongroup: **Deprecated Things are not filtered when used as a diagram root**

`Build()`'s root loop only checks `configuration.ExcludedThings.Contains(root.Iid)` before adding a root node — it never checks `IsDeprecatable`/`IsDeprecated` the way `IsNodeAllowed` does for every other reached node (see `IsNodeAllowed`, which the non-root traversal path uses and which the `VerifyThatDeprecatedThingsAreExcluded` test covers).

Failure scenario: a deprecated `RequirementsSpecification` is dropped onto the canvas as an explicit root, or matches `SelectedRootClassKinds`/`SelectedRootCategories` (auto-populated roots via `QueryRootsByClassKindAndCategory`, which also has no deprecation filter). It renders on the diagram as a root even though every other deprecated Thing reachable via traversal is correctly excluded — an inconsistency with the documented/tested behavior.

CDP4DiagramEditor/ViewModels/RelationshipTraceabilityViewModel.cs
```diff
@@ -0,0 +1,1039 @@
...
+
+        /// <summary>
+        /// Asserts whether a dragged <see cref="Thing"/> may become a root: it must belong to the iteration this panel
+        /// shows and not be a root already
+        /// </summary>
+        /// <param name="thing">The dragged <see cref="Thing"/></param>
+        /// <returns>true when the <see cref="Thing"/> may become a root</returns>
+        private bool IsDroppableRoot(Thing thing)
+        {
+            return thing.GetContainerOfType<Iteration>()?.Iid == this.Thing.Iid && this.RootThings.All(x => x.Iid != thing.Iid);
```
samatstariongroup, line 588: **`IsDroppableRoot` does not check `ExcludedThings`, so an excluded Thing can be re-added as an invisible root**

`DragOver`/`Drop` only check `RootThings.All(x => x.Iid != thing.Iid)` and the container iteration, never `ExcludedThings`.

Failure scenario: a node is excluded via the diagram context menu (added to `ExcludedThings`), then the same `Thing` is dragged from a browser back onto the canvas. `DragOver` shows the Copy drop cursor implying success, `Drop` adds it to `RootThings`, but `RelationshipGraphBuilder.Build()` skips any root whose Iid is in `configuration.ExcludedThings` — so nothing renders. The user gets a misleading accepted drop cursor with no visible result and no explanation, and `RootThings` silently accumulates a root that can never display until the user separately restores it from the Excluded list.

CDP4DiagramEditor/Behaviors/TraceabilityDiagramBehavior.cs
```diff
@@ -0,0 +1,256 @@
...
+            base.OnAttached();
+            this.AssociatedObject.DataContextChanged += this.OnDataContextChanged;
+            this.AssociatedObject.ItemsChanged += this.OnItemsChanged;
+            this.OnDataContextChanged(this.AssociatedObject, default);
+        }
+
+        /// <summary>
+        /// Unsubscribes event handlers when detaching
+        /// </summary>
+        protected override void OnDetaching()
```
samatstariongroup, line 74: **Pending `Dispatcher.BeginInvoke` layout callback is not cancelled on detach**

`OnItemsChanged` schedules `this.AssociatedObject.Dispatcher.BeginInvoke(... this.ApplyLayout() ...)` at `DispatcherPriority.Background` and only guards re-entrancy with `layoutPending`. `OnDetaching` (below) unsubscribes the `ItemsChanged`/`DataContextChanged` handlers and clears `viewModel.Behavior`, but never cancels this deferred dispatcher operation.

Failure scenario: the diagram items change (e.g. `ComputeGraph()` repopulates `Nodes`/`Edges`), scheduling a background layout pass; the user closes the traceability panel before that dispatcher callback runs. The queued callback still fires afterward and calls `this.ApplyLayout()`, which dereferences `this.AssociatedObject.Items` — if the behavior's `AssociatedObject` has already been cleared by detach, this throws a `NullReferenceException` on the UI thread.

CDP4DiagramEditor/ViewModels/RelationshipTraceabilityViewModel.cs
```diff
@@ -0,0 +1,1039 @@
...
+                    }
+                });
+        }
+
+        /// <summary>
+        /// Adds the message bus subscriptions that keep the graph in sync with the model
+        /// </summary>
+        private void AddTraceabilitySubscriptions()
+        {
+            foreach (var relationshipType in new[] { typeof(BinaryRelationship), typeof(MultiRelationship) })
```
samatstariongroup, line 859: **No message-bus subscription for property changes on already-displayed nodes**

`AddTraceabilitySubscriptions` only listens for `ObjectChangedEvent` on `BinaryRelationship`, `MultiRelationship` and `Category`. `TraceabilityNodeViewModel` is a plain, non-reactive snapshot (`Name`, `ShortName`, `Level` are captured once in its constructor), and the graph is otherwise only recomputed by these three subscriptions plus explicit user actions.

Failure scenario: a `RequirementsSpecification` (or any other Thing) already shown as a node on the diagram is renamed, or is marked `IsDeprecated = true`, in another panel while the traceability panel is open. `RelationshipGraphBuilder.IsNodeAllowed` is supposed to exclude deprecated things, and node labels should reflect the current name, but neither happens here because no event triggers `ComputeGraph()` — the diagram silently goes stale until an unrelated relationship or category change happens to force a recompute.

CDP4DiagramEditor/Helpers/RelationshipGraphBuilder.cs
```diff
@@ -0,0 +1,341 @@
...
+                        }
+
+                        if (!IsNodeAllowed(traversal.RelatedThing, configuration))
+                        {
+                            continue;
+                        }
+
+                        if (!nodes.ContainsKey(traversal.RelatedThing.Iid))
+                        {
+                            if (nodes.Count >= configuration.MaxNodes)
```
samatstariongroup, line 217: **Traversal keeps scanning every remaining level after the node cap is hit**

Once `nodes.Count >= configuration.MaxNodes` is reached for a newly-discovered thing, the code sets `maxNodeCountReached = true` and `continue`s — but the outer `for (var level = 1; level <= depth && currentLevelThings.Any(); level++)` loop is not broken. Things that were already nodes before the cap tripped still get re-visited (`nodes.ContainsKey` is true for them, bypassing the cap check) and keep populating `nextLevelThings`, so the BFS keeps walking relationships level-by-level all the way to `depth` (UI allows up to 20) purely to discard the results.

Failure scenario: a dense model with `MaxNodes` set low (e.g. the default-adjacent 300) and `DepthDown`/`DepthUp` set high hits the cap at level 2 but still performs the full relationship-lookup work for levels 3-20 on every `ComputeGraph()` call (which fires on nearly every UI property change), wasting CPU on work whose output is thrown away.

CDP4DiagramEditor/Helpers/RelationshipGraphBuilder.cs
rdevoogtatstariongroup: Fixed: extracted the deprecation check into an IsDeprecated helper and applied it in Build()'s root loop alongside the exclusion check, so both dropped roots and QueryRootsByClassKindAndCategory roots are now filtered at the single choke point

CDP4DiagramEditor/ViewModels/RelationshipTraceabilityViewModel.cs
rdevoogtatstariongroup: Fixed: rather than rejecting the drop, Drop now lifts the exclusion for that Thing before adding it as a root, since dragging something in is an explicit request to see it and the Copy cursor now tells the truth

CDP4DiagramEditor/ViewModels/RelationshipTraceabilityViewModel.cs
rdevoogtatstariongroup: Fixed: ComputeGraph now (re)subscribes to ObjectChangedEvent per rendered node Thing through a CompositeDisposable that is registered on Disposables, so renaming or deprecating a displayed Thing refreshes the diagram

CDP4DiagramEditor/Behaviors/TraceabilityDiagramBehavior.cs
rdevoogtatstariongroup: Fixed: the layoutPending bool is replaced by the DispatcherOperation itself, which OnDetaching aborts, and ApplyLayout additionally guards against a null AssociatedObject

CDP4DiagramEditor/Helpers/RelationshipGraphBuilder.cs
rdevoogtatstariongroup: Fixed: the level loop now breaks once the cap has tripped, after finishing the current level so edges between already-shown nodes are still added

CDP4DiagramEditor/ViewModels/TraceabilityEdgeViewModel.cs
```diff
@@ -0,0 +1,112 @@
...
+        public TraceabilityEdgeViewModel(RelationshipGraphEdge edge, IReadOnlyDictionary<Guid, int> nodeLevels)
+        {
+            this.Relationship = edge.Relationship;
+            this.IsUndirected = edge.Relationship is MultiRelationship;
+            this.Label = string.Join(", ", edge.Relationship.Category.Select(x => x.ShortName));
+
+            nodeLevels.TryGetValue(edge.Source.Iid, out var sourceLevel);
+            nodeLevels.TryGetValue(edge.Target.Iid, out var targetLevel);
+
+            if (sourceLevel <= targetLevel)
```
samatstariongroup, line 64: This compares raw signed `Level` values, but the class doc comment above (lines 38-41) defines ordering as "shallower to deeper traversal level" — per `RelationshipGraphNode.Level`'s own doc, 0 is the root, sign indicates direction (up/down), and magnitude is distance from root. Comparing signed values instead of magnitude means `sourceLevel <= targetLevel` is wrong for upward-only chains.

Concretely: with `DepthUp >= 2`, an edge between a node at level `-1` (1 hop up, shallower) and one at level `-2` (2 hops up, deeper) evaluates `-1 <= -2` as `false`, so the *deeper* ancestor gets picked as `FromId` and `IsReversed` is set backwards relative to the documented contract. That flows into `TraceabilityDiagramBehavior`'s Sugiyama layout (stacks the farther ancestor closer to the root than its immediate parent) and into the SVG export (arrow drawn reversed).

Suggest comparing by `Math.Abs(sourceLevel)` vs `Math.Abs(targetLevel)` (or otherwise normalizing to traversal distance) instead of the raw signed values.

CDP4DiagramEditor/ViewModels/RelationshipTraceabilityViewModel.cs
```diff
@@ -0,0 +1,1078 @@
...
+        /// </summary>
+        /// <returns>The <see cref="RelationshipGraphConfiguration"/></returns>
+        private RelationshipGraphConfiguration BuildConfiguration()
+        {
+            var configuration = new RelationshipGraphConfiguration
+            {
+                DepthDown = this.DepthDown,
+                DepthUp = this.DepthUp,
+                MaxNodes = this.MaxNodes,
+                DefaultLevelFilter = BuildFilter(this.SelectedDefaultLevelCategories, this.SelectedDefaultDirection.Direction)
```
samatstariongroup, line 665: `RelationshipGraphConfiguration.NodeFilter` is never set here. `RelationshipGraphBuilder.IsNodeAllowed` (and its dedicated unit test) fully implement and validate a ClassKind/Category filter on which reached `Thing`s become graph nodes — but with `NodeFilter` always `null`, that filter can never do anything through the UI.

Also: I don't see any UI control bound to populate a node-type filter for the panel — `SelectedRootClassKinds`/`SelectedRootCategories` only constrain which things can be *roots* (via `QueryRootsByClassKindAndCategory`), not which node types are allowed once traversal reaches deeper levels.

Worth a quick check: was node-level filtering meant to be exposed in this PR and the UI wiring got dropped, or is `NodeFilter` speculative engine capability for a later iteration? If the former, this configuration should be wired up here; if the latter, worth a comment noting it's intentionally unused for now.

samatstariongroup (commented):
Automated review pass on this feature. Left two inline comments on issues I'd want addressed before merge; filed the remaining lower-severity findings (perf/reuse cleanups) as a separate follow-up issue rather than blocking this PR on them.

CDP4DiagramEditor/ViewModels/TraceabilityEdgeViewModel.cs
rdevoogtatstariongroup: Left it as it is. The doc comment being wrong, but the code was right: the signed Level is the layout rank, so sourceLevel <= targetLevel is what stacks the upward cone above the root and the downward cone below it, whereas comparing magnitudes puts ancestors below the root.

CDP4DiagramEditor/ViewModels/RelationshipTraceabilityViewModel.cs
rdevoogtatstariongroup: Removed it. It was left over code from another implementation. that did not work good and was removed.

PR #1468, opened 2026-08-18
Feat/gh1465 Add a Verification and Validation plugin (CDP4VandV)

### LINKED ISSUE #1465: [Feature] As A User I want to create a V&V Plan, to generate V&V deliverables and plan V&V activities against requirements
I'd like to add a V&V (Verification and Validation) capability to the IME: a plugin that lets a team
plan verification activities against requirements, record their execution, and generate the standard
V&V deliverables (V&V Plan, VCD, RVM/VCRM, procedures, execution records, NCR list).

rdevoogtatstariongroup:
### Prerequisites

### Description

Closes #1465.

Adds a new plugin that turns an engineering model into a working **Verification Control Document**: who verifies
which requirement, how, at which stage gate, whether it passed, whether the design actually complies, and whether
the customer has accepted it.

### No metamodel change

Everything is expressed in stock ECSS-E-TM-10-25A concepts, so a model produced here opens in any other COMET
client without that client knowing the plugin exists. No SDK version bump, no server change, no migration.

| Concept | Encoding |
|---|---|
| V&V item | `Requirement` categorized `VnVItem`, in a `VNV` specification created on first use |
| Its attributes | `SimpleParameterValue`s keyed by parameter type short-name |
| What it verifies | `BinaryRelationship` categorized `verifies` or `validates` |
| What it measures | relationships categorized `coversParameter`, `verifiedOn`, `coversOption`, `coversState` |
| Its procedure | one `Requirement` per step, categorized `VnVStep`, linked by `hasStep` |
| Non-conformances | `ReviewItemDiscrepancy`, `RequestForDeviation`, `RequestForWaiver` |

### What it does

- **Self-seeding reference data.** "Set up V&V" checks the chained RDL, reports exactly what is missing and creates
  only that. Stage gates are defined by the user during set-up in an ordered table, because they differ per
  project; the rest is seeded from a manifest. Works on any existing model with no prior RDL curation.
- **One dialog per activity.** Right-click a requirement, fill one form, and the item, its attributes, the
  traceability link, the V&V specification and any coverage links are written in a single transaction.
- **Register browser** mirroring the stock Requirements browser grouping, with status roll-up per requirement,
  group and specification.
- **Five views** (Register, Planning, Procedure, Execution, Compliance) and a stage-gate filter, as column sets on
  one tree so selection, expansion and every context-menu action behave identically in all of them.
- **Verification procedures** as ordered steps with action, expected result, as-run result and outcome, per
  ECSS-E-ST-10-03, shown nested under their item in the tree.
- **Compliance and close-out** kept distinct from execution status, as ECSS-E-ST-10-02 Annex B requires.
- **Analysis check** that evaluates a covered parameter against the requirement's parametric constraints, honouring
  the option and state slice, and re-runs when a design value moves. It reports; it never sets a status.
- **Coverage matrix (RVM)** panel, live, with stage gates read from the RDL.
- **Excel export**: VCD, RVM, Execution Records, Procedures and NCRs.
- **Two built-in rules**: requirements nobody verifies, and V&V items unfit to execute or report.
- **RIDs, deviations, waivers, change requests and model notes**, with reply and status transitions.

### Changes outside the new plugin

One file: `CDP4-CE.sln`, registering the two new projects.

The plugin references only `CDP4Composition` and the SDK packages, and its ribbon groups target the shell-hosted
Model ribbon region, so it loads and works with any combination of other plugins present. No existing behaviour is
modified and no auto-generated file is touched.

### Notes for reviewers

Two things the IME does today that this plugin works around, both worth knowing independently of this PR:

- The IME registers **no `ThingDialog` for any annotation kind** (RID, RFD, RFW, change request, model note), so
  every browser's stock create-annotation menu items throw "not registered with the Application" and silently do
  nothing. The plugin writes them itself. `AnnotationCreator`, `AnnotationQuery`, `AnnotationKind` and
  `AnnotationDialogViewModel` are deliberately not V&V specific and would be a clean lift into `CDP4Composition`.
- The stock floating annotation window's Send button is gated on a permission evaluated once when the window
  opens, which leaves it dead with no explanation. The plugin's own Reply action reports a real error instead.


CDP4VandV/Rules/VnVItemCompletenessRule.cs
```diff
@@ -0,0 +1,191 @@
...
+    /// evidence of anything.
+    /// </summary>
+    [BuiltInRuleMetaDataExport("STARION", "VnVItemCompleteness", "A rule that flags V&V items that are not fit to be executed or reported: no traceability link, no method, no stage gate, no acceptance criteria, or a closed status without a recorded result")]
+    public class VnVItemCompletenessRule : BuiltInRule
+    {
+        /// <summary>
+        /// The statuses (see <see cref="Rdl.VandVRdlManifest"/>) that assert an outcome, and therefore require a
+        /// recorded result.
+        /// </summary>
+        private static readonly string[] ConcludedStatuses = { "Passed", "Failed", "Closed" };
```
samatstariongroup: `ConcludedStatuses` includes `"Closed"`, which is never a valid `vnv_status` value (close-out is tracked separately via the `vnv_closed` boolean), and it omits `Waived`/`Deviated`/`Not Applicable`.

**Failure scenario:** A V&V item concluded as Waived, Deviated, or Not Applicable (or closed out via `vnv_closed`) with `vnv_result` left blank is never flagged by the 'concluded status without a recorded result' check, silently letting an under-evidenced closure pass the audit this rule exists to enforce.

CDP4VandV/ViewModels/VandVItemDialogViewModel.cs
```diff
@@ -0,0 +1,1329 @@
...
+        /// Pre-selects the coverage of an existing V&amp;V item so the Coverage tab opens on what is already linked.
+        /// </summary>
+        /// <param name="iteration">The iteration.</param>
+        /// <param name="vandVItem">The V&amp;V item being edited.</param>
+        private void LoadCoverage(Iteration iteration, Requirement vandVItem)
+        {
+            var parameter = VandVCoverageWriter.QueryCoveredThings<ParameterOrOverrideBase>(iteration, vandVItem, VandVCoverageWriter.CoversParameter).FirstOrDefault();
+            var element = VandVCoverageWriter.QueryCoveredThings<ElementDefinition>(iteration, vandVItem, VandVCoverageWriter.VerifiedOn).FirstOrDefault();
+
+            this.SelectedElementDefinition = element ?? parameter?.Container as ElementDefinition;
```
samatstariongroup: `LoadCoverage` (and `PreselectParameter` around line 1077) resolves the owning element via `parameter?.Container as ElementDefinition`, which is `null` whenever the parameter is a `ParameterOverride` (its `Container` is an `ElementUsage`, not an `ElementDefinition`). Compare with the correct fallback already used in `VandVBrowserViewModel.cs` (`parameter.Container as ElementDefinition ?? (parameter.Container as ElementUsage)?.ElementDefinition`).

**Failure scenario:** Editing a V&V item whose stored coverage points at an option/state-dependent `ParameterOverride` leaves `SelectedElementDefinition` null, so the Coverage tab shows an empty element picker even though the item holds that override.

Requirements/ViewModels/VandV/VandVItemDialogViewModel.cs
```diff
@@ -0,0 +1,1329 @@
...
+                this.SelectedOptions.Select(x => x.Iid).ToList(),
+                this.SelectedStates.Select(x => x.Iid).ToList());
+
+            this.AnalysisCheck = result.Display;
+        }
+
+        /// <summary>
+        /// Refreshes the parameter list when the element definition changes.
+        /// </summary>
+        private void PopulateParameters()
```
samatstariongroup, line 1182: `PopulateParameters` re-adds and re-selects the previously selected parameter even after `SelectedElementDefinition` has changed to a different element, instead of clearing a selection that no longer belongs to the new element.

**Failure scenario:** User selects Element A, picks Parameter P, then switches the Coverage tab's element combo to unrelated Element B; `SelectedParameter` stays P (still belonging to A) injected into B's parameter list, so a mismatched element/parameter pairing can be saved as the item's coverage.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
```diff
@@ -0,0 +1,1362 @@
...
+        /// </summary>
+        /// <param name="sender">The row the parameter was dropped on.</param>
+        /// <param name="parameter">The dropped parameter.</param>
+        /// <remarks>
+        /// A plain parameter is coupled outright, there is nothing to choose. An option- or state-dependent parameter
+        /// is ambiguous: the drag payload is the parameter itself, never the option/state row it was started from, so
+        /// the browser cannot know which slice was meant. Guessing "all of them" is wrong more often than it is right,
+        /// so the dialog opens on the Coverage tab with the parameter already selected and nothing ticked.
+        /// </remarks>
+        private async void OnParameterDropped(object sender, ParameterOrOverrideBase parameter)
```
samatstariongroup: `OnParameterDropped` is `async void`, but only the `SetCoverageAsync` call (~line 646) is wrapped in try/catch. The preceding `QueryCoverageRelationships` call (~line 626) and the `DXMessageBox.Show` confirmation dialog (~line 628) are unguarded.

**Failure scenario:** If `QueryCoverageRelationships` or the confirmation dialog throws, the exception escapes an `async void` event handler and crashes the application instead of showing the same 'could not be coupled' error message used for the guarded portion.

samatstariongroup (commented):
Automated code review findings (Claude Code).

Requirements/Rdl/VandVRdlService.cs
```diff
@@ -0,0 +1,341 @@
...
+                rule.ParameterType.AddRange(
+                    definition.ParameterTypeShortNames
+                        .Select(shortName => ResolveParameterType(shortName, createdParameterTypes, mrdl))
+                        .Where(parameterType => parameterType != null));
+
+                clone.Rule.Add(rule);
+                transaction.Create(rule);
+            }
+
+            if (missing.RequirementTargetCategory != null)
```
samatstariongroup, line 252: `Seed()` only creates the verifies/validates `BinaryRelationshipRule`s when `RequirementTargetCategory` resolves; when it's null (no matching requirement category found in the RDL chain), the rules are skipped with no warning to the user.

**Failure scenario:** In a model with non-standard requirements category naming, 'Set up V&V' reports success ('reference data has been created') but the verifies/validates rules are missing; creating a V&V item afterward throws "The 'verifies' category was not found. Run 'Set up V&V' first" even though the user just ran it.

Requirements/Services/VandVCoverageWriter.cs
```diff
@@ -0,0 +1,203 @@
...
+        /// <param name="iterationClone">The cloned iteration the relationship is added to.</param>
+        /// <param name="transaction">The transaction.</param>
+        /// <param name="mrdl">The model reference data library.</param>
+        /// <param name="source">The V&amp;V item.</param>
+        /// <param name="target">The covered thing.</param>
+        /// <param name="categoryShortName">The coverage category short-name.</param>
+        /// <param name="owner">The owner of the relationship.</param>
+        private static void AddLink(Iteration iterationClone, IThingTransaction transaction, ReferenceDataLibrary mrdl, Thing source, Thing target, string categoryShortName, DomainOfExpertise owner)
+        {
+            var category = mrdl.QueryCategoriesFromChainOfRdls().FirstOrDefault(x => x.ShortName == categoryShortName);
```
samatstariongroup, line 185: `AddLink` silently returns without creating the relationship or reporting any error when the coverage category short-name can't be resolved in the RDL chain, unlike the analogous `ResolveCategory` helpers elsewhere which throw.

**Failure scenario:** `SetCoverageAsync` runs on a model whose RDL chain is missing e.g. the `coversParameter` category (partially-run setup, or a category later removed); the write completes 'successfully' but the requested coverage link is silently dropped with no indication to the user.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
```diff
@@ -0,0 +1,1362 @@
...
+        /// <summary>
+        /// Reacts to a requirement change. A V&amp;V item only ever affects the coverage children, so the tree is left
+        /// standing, rebuilding it wholesale would collapse everything the user had expanded. The tree is only
+        /// rebuilt when the shape actually changed: a requirement was added, removed, deprecated or re-grouped.
+        /// </summary>
+        /// <param name="requirement">The changed requirement.</param>
+        /// <param name="eventKind">The kind of change.</param>
+        private void OnRequirementChanged(Requirement requirement, EventKind eventKind)
+        {
+            if (eventKind == EventKind.Removed)
```
samatstariongroup, line 1083: `OnRequirementChanged`'s `EventKind.Removed` branch runs before the iteration-membership filter (`BelongsToThisIteration`) that every other event handler in this class applies, so it reacts to `Requirement` deletions from any open model/iteration.

**Failure scenario:** With two V&V browser panels open for different models, deleting a requirement in one model triggers `RefreshCoverage()` in the other panel too, resetting its `SelectedThing` and disrupting the user's selection in a panel unrelated to the edit.

CDP4VandV/Services/VandVItemCreator.cs
```diff
@@ -0,0 +1,395 @@
...
+                throw new ArgumentNullException(nameof(requirement));
+            }
+
+            if (attributes == null)
+            {
+                throw new ArgumentNullException(nameof(attributes));
+            }
+
+            var iteration = requirement.GetContainerOfType<Iteration>();
+            var mrdl = ((EngineeringModel)iteration.Container).EngineeringModelSetup.RequiredRdl.Single();
```
samatstariongroup: Write paths (`CreateAsync`/`UpdateAsync`, and similarly `VandVCoverageWriter.SetCoverageAsync`/`VandVProcedureWriter.WriteAsync`) resolve `EngineeringModelSetup.RequiredRdl` via `.Single()`, while the corresponding read-side gate (`CanCreate`, `QueryStages`) uses `.FirstOrDefault()`, which tolerates zero entries.

**Failure scenario:** If `RequiredRdl` has zero or more than one entry when a write path runs (a state not caught by the more permissive `CanCreate` check), `.Single()` throws an unhandled `InvalidOperationException` that surfaces as a generic failure instead of the clear 'Run Set up V&V first' guidance used elsewhere.

Requirements/ViewModels/VandV/VandVModelMenuItemViewModel.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// </summary>
+        public ReactiveCommand<Unit, Unit> SetUpVnVCommand { get; }
+
+        /// <summary>
+        /// Checks the model's chained reference data, and seeds the missing V&amp;V manifest items after confirmation.
+        /// </summary>
+        /// <returns>A <see cref="Task"/>.</returns>
+        private async Task ExecuteSetUpVnV()
+        {
+            var mrdl = this.model.EngineeringModelSetup.RequiredRdl.FirstOrDefault();
```
samatstariongroup, line 124: `ExecuteSetUpVnV` resolves `EngineeringModelSetup.RequiredRdl.FirstOrDefault()` and, if null, silently returns with no message box or log.

**Failure scenario:** A user opens an incompletely configured model and clicks 'Set up V&V' from the Model ribbon; nothing happens at all, no error or explanation, making the feature appear silently broken.

CDP4VandV/ViewModels/VandVBrowserViewModel.cs
```diff
@@ -0,0 +1,1362 @@
...
+        {
+            if (!(this.SelectedThing is RequirementCoverageRowViewModel row))
+            {
+                return;
+            }
+
+            if (!VandVItemCreator.CanCreate(this.Thing))
+            {
+                DXMessageBox.Show(
+                    "The V&V reference data is not present in this model yet. Run 'Set up V&V' on the Requirements ribbon first.",
```
samatstariongroup: The warning shown when creating a V&V item before RDL setup still says to run 'Set up V&V' on the 'Requirements ribbon', but this same PR (e31ab962) moved the V&V ribbon entry to `RegionNames.ModelRibbonPageRegion` and removed the Requirements-page region entirely (README was updated but this string was not).

**Failure scenario:** A user without V&V reference data tries to create a V&V item, is told to look on the 'Requirements ribbon', but 'Set up V&V' is actually under the Model ribbon tab now, leaving the user unable to find the feature.

CDP4VandV/Services/VandVCoverageQuery.cs
```diff
@@ -0,0 +1,331 @@
...
+        /// <returns>The cell text, or an empty string when nothing is planned at that stage.</returns>
+        public string CellText(string stage)
+        {
+            var atStage = this.VandVItems
+                .Where(item => VandVCoverageQuery.AreSameEnumValue(VandVCoverageQuery.Attribute(item, "vnv_stage"), stage))
+                .Select(item =>
+                {
+                    var method = VandVCoverageQuery.Attribute(item, "vnv_method");
+                    var status = VandVCoverageQuery.Attribute(item, "vnv_status");
+                    var activity = string.IsNullOrWhiteSpace(status) ? method : $"{method} ({status})";
```
samatstariongroup: `CellText` builds `activity = $"{method} ({status})"`, which produces a malformed cell like " (Executed)" with a leading space and orphan parentheses when a V&V item has a status but no method attribute recorded.

**Failure scenario:** A V&V item has `vnv_status` set (e.g. 'Executed') but `vnv_method` left blank; the RVM matrix cell shows "ItemShortName:  (Executed)" instead of a clean "ItemShortName: Executed".

CDP4VandV/Rules/VnVItemCompletenessRule.cs
rdevoogtatstariongroup: Fixed: the list is now Passed, Failed, Waived, Deviated and Not Applicable, the stray "Closed" entry is gone, and an item closed out via vnv_closed without a recorded result is now flagged as well.

CDP4VandV/ViewModels/VandVItemDialogViewModel.cs
rdevoogtatstariongroup: Fixed: LoadCoverage and PreselectParameter now resolve the element through a shared helper that falls back to the ElementUsage's ElementDefinition for overrides.

Requirements/ViewModels/VandV/VandVItemDialogViewModel.cs
rdevoogtatstariongroup: Fixed: the previous selection is only kept when it belongs to the newly selected element (directly or via a usage), so a mismatched element and parameter pairing can no longer be saved.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
rdevoogtatstariongroup: Fixed: the whole handler body, including the coverage query and the confirmation dialog, is now inside the try/catch that shows the coupling error.

Requirements/Services/VandVCoverageWriter.cs
rdevoogtatstariongroup: Fixed: a missing coverage category now throws the same "Run 'Set up V&V' first" InvalidOperationException the other resolvers use, which the calling browsers catch and display.

Requirements/Rdl/VandVRdlService.cs
rdevoogtatstariongroup: Fixed: when no requirement category resolves, the success message now states that the verifies/validates rules were skipped and tells the user to create a REQ/REQUIREMENT category and rerun the set-up.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
rdevoogtatstariongroup: Fixed: the Removed branch now only reacts when this browser's own rows reference the deleted requirement, so deletions in other open models no longer refresh the panel or disturb its selection.

Requirements/ViewModels/VandV/VandVModelMenuItemViewModel.cs
rdevoogtatstariongroup: Fixed: a model without a required reference data library now gets a message box explaining that a site administrator must attach one, instead of nothing happening.

CDP4VandV/Services/VandVItemCreator.cs
rdevoogtatstariongroup: Fixed: all write paths now resolve the RDL through a shared VandVCoverageQuery.QueryRequiredRdl helper that throws the actionable "Run 'Set up V&V' first" message instead of a bare Single() exception.

CDP4VandV/ViewModels/VandVBrowserViewModel.cs
rdevoogtatstariongroup: Fixed: the message now points at the Requirements ribbon tab, which is where Set up V&V lives again after the feature was merged into the Requirements plugin per this review.

CDP4VandV/Services/VandVCoverageQuery.cs
rdevoogtatstariongroup: Fixed: a status-only item now renders as "ItemShortName: Executed" via an explicit three-way branch over method and status.

Requirements/Services/VandVItemCreator.cs
```diff
@@ -0,0 +1,395 @@
...
+        /// The short-name of the requirements specification the V&amp;V items live in.
+        /// </summary>
+        public const string VandVSpecificationShortName = "VNV";
+
+        /// <summary>
+        /// Asserts whether the model has the reference data required to create V&amp;V items (i.e. "Set up V&amp;V" has run).
+        /// </summary>
+        /// <param name="iteration">The <see cref="Iteration"/> the item would be created in.</param>
+        /// <returns>true when the <c>VnV Item</c> and <c>verifies</c> categories and the V&amp;V parameter types exist.</returns>
+        public static bool CanCreate(Iteration iteration)
```
samatstariongroup, line 73: `CanCreate` only validates the VnVItem/verifies categories and the `vnv_method` parameter type, so create/edit can proceed even when coverage or procedure-step categories are missing from the RDL, allowing a partial commit.

**Failure scenario:** On a partially-seeded RDL, `ExecuteCreateVandVItem` gates only on `CanCreate`, then awaits `CreateAsync` (commits the item), then `SaveCoverageAsync` and `procedureWriter.WriteAsync`, which throw `InvalidOperationException` for missing `coversParameter`/`verifiedOn`/`coversOption`/`coversState` or `VnVStep`/`hasStep` categories. The single catch block shows a generic 'could not be created' message even though the item was already persisted, leaving an orphaned V&V item the user believes never happened. `EditVandVItem` doesn't call `CanCreate` at all, so a simple rename can silently leave coverage/procedure writes half-applied while reporting 'could not be updated'.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
```diff
@@ -0,0 +1,1366 @@
...
+        /// <param name="pluginSettingsService">The <see cref="IPluginSettingsService"/>.</param>
+        public VandVBrowserViewModel(Iteration iteration, ISession session, IThingDialogNavigationService thingDialogNavigationService, IPanelNavigationService panelNavigationService, IDialogNavigationService dialogNavigationService, IPluginSettingsService pluginSettingsService)
+            : base(iteration, session, thingDialogNavigationService, panelNavigationService, dialogNavigationService, pluginSettingsService)
+        {
+            this.Caption = $"{PanelCaption}, iteration_{this.Thing.IterationSetup.IterationNumber}";
+            this.ToolTip = $"{((EngineeringModel)this.Thing.Container).EngineeringModelSetup.Name}\n{this.Thing.IDalUri}\n{this.Session.ActivePerson.Name}";
+
+            this.SpecificationRows = new DisposableReactiveList<VandVSpecificationRowViewModel>();
+
+            this.CreateVandVItemCommand = ReactiveCommandCreator.CreateAsyncTask(
```
samatstariongroup, line 169: `VandVBrowserViewModel` never overrides `ComputePermission()` or calls `PermissionService.CanWrite`, so V&V item creation, annotation creation, and reply/status-change commands are gated only on selection state, not write permission.

**Failure scenario:** A read-only or under-permissioned participant sees 'Create V&V Item', 'Raise RID/RFD/RFW', and reply/close-annotation actions fully enabled (unlike the sibling `RequirementsBrowserViewModel`, which gates equivalent commands on `PermissionService.CanWrite`). They fill out a full dialog and only discover the denial as a raw exception after `session.Write` round-trips to the server, instead of the command being disabled up front.

Requirements/Rdl/VandVRdlService.cs
```diff
@@ -0,0 +1,341 @@
...
+        /// </summary>
+        /// <param name="session">The <see cref="ISession"/> used to write the transaction.</param>
+        /// <param name="mrdl">The target <see cref="ModelReferenceDataLibrary"/>; self-contained and travels with the model.</param>
+        /// <param name="missing">The <see cref="VandVRdlCheckResult"/> produced by <see cref="Check"/>.</param>
+        /// <param name="stageGates">
+        /// The project's stage gates, replacing the manifest's example list when supplied. Stage gates differ per
+        /// project, so the set-up flow asks for them rather than silently seeding a hard-coded set.
+        /// </param>
+        /// <returns>A <see cref="Task"/> that completes when the write has been dispatched.</returns>
+        public async Task Seed(ISession session, ModelReferenceDataLibrary mrdl, VandVRdlCheckResult missing, IReadOnlyList<string> stageGates = null)
```
samatstariongroup, line 161: `VandVRdlService.Seed` uses a stale 'missing categories' snapshot computed before user-interaction delays (confirmation dialog, stage-gate dialog) with no re-check immediately before writing, and no short-name uniqueness guard.

**Failure scenario:** Two users open 'Set up V&V' concurrently on the same model. Both `Check()` calls see the same missing RDL elements, both confirm through the dialogs, and both `Seed()` calls create their own `ParameterType`/`Category`/`Rule` objects with new GUIDs but identical ShortNames, permanently duplicating V&V reference data since nothing enforces short-name uniqueness at write time.

Requirements/ViewModels/VandV/VandVItemDialogViewModel.cs
```diff
@@ -0,0 +1,1342 @@
...
+        /// Pre-selects the coverage of an existing V&amp;V item so the Coverage tab opens on what is already linked.
+        /// </summary>
+        /// <param name="iteration">The iteration.</param>
+        /// <param name="vandVItem">The V&amp;V item being edited.</param>
+        private void LoadCoverage(Iteration iteration, Requirement vandVItem)
+        {
+            var parameter = VandVCoverageWriter.QueryCoveredThings<ParameterOrOverrideBase>(iteration, vandVItem, VandVCoverageWriter.CoversParameter).FirstOrDefault();
+            var element = VandVCoverageWriter.QueryCoveredThings<ElementDefinition>(iteration, vandVItem, VandVCoverageWriter.VerifiedOn).FirstOrDefault();
+
+            this.SelectedElementDefinition = element ?? QueryOwningElement(parameter);
```
samatstariongroup: `LoadCoverage`/`PreselectParameter` assign `SelectedElementDefinition` before `SelectedParameter`, but `PopulateParameters` (fired synchronously off `SelectedElementDefinition`) reads the still-stale `SelectedParameter`, so the fix-up that keeps a `ParameterOverride` visible in the list never runs.

**Failure scenario:** Editing a V&V item whose covered parameter is an option/state-dependent `ParameterOverride`: `PopulateParameters` builds `PossibleParameters` from `SelectedElementDefinition.Parameter` only (never `ParameterOverride`), then `SelectedParameter` is set afterward to a value absent from that list. The 'Covers Parameter' ComboBoxEdit renders blank/unselected even though the underlying saved value is correct, misleading the user into thinking coverage was lost.

Requirements/Services/VandVItemCreator.cs
```diff
@@ -0,0 +1,395 @@
...
+        /// <param name="iteration">The iteration.</param>
+        /// <param name="vandVItem">The V&amp;V item.</param>
+        /// <returns>The relationship, or null when the item is not linked.</returns>
+        public static BinaryRelationship QueryCoveringRelationship(Iteration iteration, Requirement vandVItem)
+        {
+            return iteration.Relationship
+                .OfType<BinaryRelationship>()
+                .FirstOrDefault(x =>
+                    x.Source == vandVItem
+                    && x.Category.Any(category => category.ShortName == "verifies" || category.ShortName == "validates"));
```
samatstariongroup: Category short-name literals ('VnVItem', 'verifies', 'validates', 'coversParameter', etc.) are hand-typed independently in at least 4 files (`VandVItemCreator`, `VandVCoverageWriter`, `RequirementVnVCoverageRule`, `VandVCoverageQuery`) instead of referencing a single shared constant.

**Failure scenario:** A typo in any one of the ~10+ independent occurrences compiles cleanly and silently creates a disconnected notion of coverage/V&V-item. `RequirementVnVCoverageRule`'s self-exclusion and coverage detection both depend on its private `VnVItemCategoryShortName` copy matching the manifest exactly; drift would cause every V&V item to be flagged as an uncovered requirement, polluting the shared Rule Verification results.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
```diff
@@ -0,0 +1,1366 @@
...
+            this.Disposables.Add(
+                this.CDPMessageBus.Listen<ObjectChangedEvent>(typeof(RequirementsSpecification))
+                    .Merge(this.CDPMessageBus.Listen<ObjectChangedEvent>(typeof(RequirementsGroup)))
+                    .Where(x => x.EventKind != EventKind.Updated)
+                    .ObserveOn(RxApp.MainThreadScheduler)
+                    .Subscribe(_ => this.Rebuild()));
+
+            // the analysis check reads live parameter values, so a design change has to re-run it; ValueSet changes
+            // are what move a parameter, and they never touch the V&V item itself
+            this.Disposables.Add(
```
samatstariongroup, line 1025: The `ParameterValueSet`/`ParameterOverrideValueSet` subscription filters only by iteration, so any single parameter edit anywhere triggers `RefreshAnalysisChecks()` to re-run `VandVAnalysisChecker.Check` over every V&V item in the tree, each doing multiple unindexed O(relationships) scans.

**Failure scenario:** On a model with a few hundred V&V items and thousands of relationships, editing one unrelated parameter value causes O(N items × 4 × R relationships) synchronous work on the UI thread, visibly stuttering the UI, when only the 0-1 items actually covering that parameter needed re-evaluation.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
```diff
@@ -0,0 +1,1366 @@
...
+                    .Subscribe(_ => this.RefreshAnnotationStates()));
+
+            // closing the matrix panel disposes its view-model, so the cached instance must be dropped or the next
+            // "Open Coverage Matrix" would re-dock a dead panel that no longer reacts to model changes
+            this.Disposables.Add(
+                this.CDPMessageBus.Listen<NavigationPanelEvent>()
+                    .Where(x => x.ViewModel == this.matrixViewModel && x.PanelStatus == PanelStatus.Closed)
+                    .Subscribe(_ => this.matrixViewModel = null));
+
+            this.Disposables.Add(
```
samatstariongroup, line 1054: The `BinaryRelationship` change subscription filters only by iteration, not by `VandVCoverageQuery.IsCoverageLink`, so `RefreshCoverage()` tears down and rebuilds every requirement row's V&V item children tree-wide on any relationship change.

**Failure scenario:** Creating or editing an ordinary, non-coverage `BinaryRelationship` elsewhere in the same iteration (e.g. a plain requirement trace link) disposes and recreates all V&V item rows under every requirement in the entire tree, unsubscribing/resubscribing every row's event handlers unnecessarily.

samatstariongroup (commented):
Automated code review findings (Claude Code) — updated review after the merge with the Requirements plugin.

Requirements/ViewModels/VandV/VandVMatrixViewModel.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            this.Summary = model.Coverages.Count == 0
+                ? "No requirements in this iteration."
+                : $"{model.Coverages.Count - model.UncoveredCount} of {model.Coverages.Count} requirements covered, {model.UncoveredCount} uncovered.";
+        }
+
+        /// <summary>
+        /// Rebuilds the matrix whenever a V&amp;V item, its attributes, or a traceability relationship changes.
+        /// </summary>
+        private void AddSubscriptions()
```
samatstariongroup, line 195: `VandVMatrixViewModel`'s `BinaryRelationship`/`Requirement`/`SimpleParameterValue` subscriptions have no iteration filter, unlike the equivalent `VandVBrowserViewModel` subscriptions.

**Failure scenario:** With two engineering models open, editing a `Requirement`, `BinaryRelationship`, or `SimpleParameterValue` in model B triggers a full `BuildMatrix()` rebuild of the coverage matrix panel open for unrelated model A on every such edit.

Requirements/Services/VandVCoverageQuery.cs
```diff
@@ -0,0 +1,364 @@
...
+        /// Summarises a set of V&amp;V items as "how many are done, failed and still open".
+        /// </summary>
+        /// <param name="items">The V&amp;V items being rolled up.</param>
+        /// <returns>The counts of closed-out, failed and open items.</returns>
+        /// <remarks>
+        /// Close-out wins over execution status. An item explicitly closed out counts as done whatever its execution
+        /// status says, and an item that is not closed out counts as done only when its execution status is one that
+        /// needs no further action. That ordering is what makes the tree agree with the VCD's close-out column.
+        /// </remarks>
+        public static VandVStatusRollUp RollUp(IEnumerable<Requirement> items)
```
samatstariongroup, line 256: `RollUp` counts any item with `vnv_closed=true` as 'passed' regardless of its actual status or compliance shortfall, which disagrees with `VnVItemCompletenessRule`'s violation check for shortfalls closed without an accepted concession.

**Failure scenario:** An item that failed verification or has non-compliant design, but was closed out without an accepted waiver/deviation, is counted as 'passed' in the tree/specification roll-up (showing e.g. 'N/N verified') while `VnVItemCompletenessRule` simultaneously reports a rule violation for that same item — a visible, confusing inconsistency between two parts of the same feature (documented as intentional in a code comment, but still a real discrepancy worth double-checking).

Requirements/Services/VandVItemCreator.cs
```diff
@@ -0,0 +1,395 @@
...
+            transaction.Create(simpleParameterValue);
+        }
+
+        /// <summary>
+        /// Resolves a required <see cref="Category"/> by short-name from the RDL chain.
+        /// </summary>
+        /// <param name="mrdl">The model reference data library.</param>
+        /// <param name="shortName">The category short-name.</param>
+        /// <returns>The resolved <see cref="Category"/>.</returns>
+        private static Category ResolveCategory(ReferenceDataLibrary mrdl, string shortName)
```
samatstariongroup: `ResolveCategory` is duplicated byte-for-byte in both `VandVItemCreator.cs` (~lines 383-393) and `VandVProcedureWriter.cs` (~lines 375-385).

**Failure scenario:** A future fix to the lookup logic or error message (e.g. improving the 'Run Set up V&V first' guidance) applied to one copy has no signal that a second identical copy exists in the sibling service, so the two will quietly diverge over time.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
```diff
@@ -0,0 +1,1444 @@
...
+            base.ComputePermission();
+
+            // called once from the base constructor, before this view-model's own state exists
+            if (this.Thing == null)
+            {
+                return;
+            }
+
+            this.CanCreateVandVItem = this.PermissionService.CanWrite(ClassKind.Requirement, this.Thing);
+            this.CanCreateAnnotation = this.PermissionService.CanWrite(ClassKind.RequestForWaiver, this.Thing.TopContainer);
```
samatstariongroup: **[correctness, CONFIRMED]** Single permission check gates all 5 create-item kinds

A single permission check on `ClassKind.RequestForWaiver` gates the create-command for all 5 distinct annotation kinds (ReviewItemDiscrepancy, RequestForDeviation, ChangeRequest, ModelNote, RequestForWaiver).

**Failure scenario:** A domain with write permission on ChangeRequest but not RequestForWaiver sees the 'Create a Change Request' command incorrectly enabled/disabled based only on the RequestForWaiver permission check, so it can appear enabled and then fail server-side, or appear disabled when it should be usable.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
```diff
@@ -0,0 +1,1444 @@
...
+
+            var dialogViewModel = new VandVItemDialogViewModel(row.Thing, this.Session);
+            var result = this.DialogNavigationService.NavigateModal(dialogViewModel);
+
+            if (result == null || result.Result != true)
+            {
+                return;
+            }
+
+            try
```
samatstariongroup, line 423: **[correctness, CONFIRMED]** V&V create/edit is 3 unguarded sequential server writes

Creating/editing a V&V item is three sequential, independently-committed server writes (item create, coverage write, procedure write) with no rollback on partial failure.

**Failure scenario:** `itemCreator.CreateAsync` succeeds, then `SetCoverageAsync` or `procedureWriter.WriteAsync` throws (network blip, revision conflict) — the Requirement is permanently committed on the server with no coverage/procedure, and the user only sees a generic error dialog with no indication the item was actually persisted (an orphaned V&V item).

Requirements/Services/VandVItemCreator.cs
```diff
@@ -0,0 +1,383 @@
...
+
+            if (mrdl == null)
+            {
+                return false;
+            }
+
+            var categories = new HashSet<string>(mrdl.QueryCategoriesFromChainOfRdls().Select(x => x.ShortName));
+            var parameterTypes = new HashSet<string>(mrdl.QueryParameterTypesFromChainOfRdls().Select(x => x.ShortName));
+
+            return VandVCategory.RequiredForItemWrite.All(categories.Contains)
```
samatstariongroup, line 90: **[correctness, CONFIRMED]** CanCreate only checks vnv_method, not other mandatory params

`CanCreate` only validates the presence of the `vnv_method` parameter type in the RDL, even though `vnv_stage`, `vnv_acceptance`, and `vnv_status` are also mandatory per the manifest and the dialog's own validation.

**Failure scenario:** An RDL has `vnv_method` but is missing `vnv_stage`: `CanCreate` returns true, the dialog lets the user pick a stage, but AddAttribute/SetAttribute silently no-op (return) when the ParameterType lookup fails, so the stage value the user entered is silently dropped with no error.

Requirements/ViewModels/VandV/VandVMatrixViewModel.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        /// <summary>
+        /// Rebuilds the matrix from the current model state.
+        /// </summary>
+        private void BuildMatrix()
+        {
+            var model = VandVCoverageQuery.Build(this.Thing);
+
+            var table = new DataTable();
+            table.Columns.Add(RequirementColumn, typeof(string));
```
samatstariongroup, line 146: **[correctness, CONFIRMED]** Stage-gate named 'Requirement' corrupts identity column

The 'Requirement'/'Requirement Name' DataTable columns are added unconditionally, but the stage-gate column loop silently skips adding a duplicate if a stage gate happens to share that name, then overwrites the identity column with stage-gate text.

**Failure scenario:** A user names a stage gate 'Requirement' or 'Requirement Name' in the Set-up V&V dialog (free text); every row in the coverage matrix panel then shows stage-gate coverage text instead of the requirement's short name/name, silently corrupting the identity column with no error raised.

Requirements/ViewModels/VandV/StageGateDialogViewModel.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// Gets the command that cancels the whole set-up.
+        /// </summary>
+        public ReactiveCommand<Unit, Unit> CancelCommand { get; }
+
+        /// <summary>
+        /// Reads the table into the stage gate list: trimmed, blanks dropped, duplicates removed while keeping the
+        /// order the user arranged.
+        /// </summary>
+        /// <returns>The stage gates, in order.</returns>
+        public IReadOnlyList<string> ParseStageGates()
```
samatstariongroup, line 137: **[correctness, CONFIRMED]** Stage-gate dedup misses hyphen/space ShortName collisions

Stage-gate name deduplication uses OrdinalIgnoreCase only, while the manifest's `ToShortName` normalizes both spaces and hyphens to underscore, so differently-typed names can collide into one ShortName when seeded.

**Failure scenario:** A user enters both 'In Service' and 'In-Service' as distinct stage gates; `ParseStageGates` treats them as unique, but `VandVRdlService.CreateParameterType` derives the identical ShortName 'In_Service' for both, silently creating two EnumerationValueDefinitions with colliding ShortNames and different Names (no uniqueness enforcement in CDP4Dal), corrupting the RDL.

Requirements/ViewModels/VandV/StageGateDialogViewModel.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            var hasSelection = this.WhenAnyValue(x => x.SelectedGate).Select(gate => gate != null);
+
+            this.AddGateCommand = ReactiveCommandCreator.Create(this.ExecuteAddGate);
+            this.RemoveGateCommand = ReactiveCommandCreator.Create(this.ExecuteRemoveGate, hasSelection);
+            this.MoveGateUpCommand = ReactiveCommandCreator.Create(() => this.MoveGate(-1), hasSelection);
+            this.MoveGateDownCommand = ReactiveCommandCreator.Create(() => this.MoveGate(1), hasSelection);
+
+            // the list contents drive OK, so an empty or blank-only table cannot be accepted
+            var canOk = this.StageGates.Changed
```
samatstariongroup: **[correctness, CONFIRMED]** OK button stays enabled after blanking stage-gate names

The OK command's CanExecute is derived from `StageGates.Changed` (structural changes only), not per-row Name property edits, so editing existing rows' text to blank/duplicate doesn't disable OK.

**Failure scenario:** User blanks out every stage-gate name via inline textboxes without adding/removing rows; OK stays clickable from the last Add/Remove-derived CanExecute state even though `ParseStageGates()` would now return an empty list, silently falling back to manifest defaults instead of reflecting what the user typed.

Requirements/ViewModels/VandV/Rows/VandVItemRowViewModel.cs
```diff
@@ -0,0 +1,474 @@
...
+                : System.Windows.DragDropEffects.None;
+        }
+
+        /// <summary>
+        /// Handles the drop of a parameter onto this V&amp;V item by raising <see cref="ParameterDropped"/>; the browser
+        /// owns the write so this row stays free of transaction logic.
+        /// </summary>
+        /// <param name="dropInfo">The <see cref="IDropInfo"/>.</param>
+        /// <returns>A completed <see cref="Task"/>.</returns>
+        public Task Drop(IDropInfo dropInfo)
```
samatstariongroup: **[correctness, CONFIRMED]** IsBusy cleared before async drop-handler actually finishes

`Drop()` synchronously fires an event to an async void handler and returns a completed Task immediately, so `IsBusy` is cleared before the actual confirmation dialog and coverage write finish, allowing a concurrent second drop.

**Failure scenario:** User drags a parameter onto a V&V item; `IsBusy` flips back to false right after the synchronous event fires, while `OnParameterDropped`'s confirmation dialog and `SetCoverageAsync` write are still in flight; a second drop or action can start concurrently and race the first coverage write.

Requirements/Rdl/VandVRdlManifest.cs
```diff
@@ -0,0 +1,410 @@
...
+    /// V&amp;V capability needs. The checker (<see cref="VandVRdlService.Check"/>) and the seeder
+    /// (<see cref="VandVRdlService.Seed"/>) are both plain functions over these lists, so adding an attribute later
+    /// is one list entry, not a code change in two places.
+    /// </summary>
+    public static class VandVRdlManifest
+    {
+        /// <summary>
+        /// Gets the <see cref="CDP4Common.SiteDirectoryData.ParameterType"/>s the V&amp;V capability needs.
+        /// </summary>
+        public static IReadOnlyList<VandVParameterTypeDefinition> ParameterTypes { get; } = new[]
```
samatstariongroup, line 491: **[reuse, PLAUSIBLE]** Parameter-type short-names not centralized as constants

Unlike category short-names (centralized in `VandVCategory` constants specifically to avoid typo bugs, per the file's own comment), the ~25 parameter-type short-names are never centralized — they are re-typed as raw string literals dozens of times across 8+ consumer files.

**Failure scenario:** A typo in one of the ~50 raw `"vnv_stage"`/`"vnv_status"` style literals (e.g. in VandVWorkbookExporter.cs) compiles cleanly and silently fails the ShortName lookup at runtime, causing that attribute to read/write as blank/missing with no error — the exact class of bug the VandVCategory constants were introduced to eliminate, just left unaddressed for parameter types.

samatstariongroup (commented):
Automated code review findings for this PR (posted on behalf of the requester). 7 CONFIRMED correctness issues and 3 PLAUSIBLE quality/perf issues found — see inline comments.

Requirements/Services/VandVCoverageQuery.cs
```diff
@@ -0,0 +1,418 @@
...
+        /// <param name="left">One value.</param>
+        /// <param name="right">The other value.</param>
+        /// <returns>true when they denote the same enumeration value.</returns>
+        /// <remarks>
+        /// Both underscore and hyphen normalise to a space, because <c>VandVRdlManifest.ToShortName</c> maps both to
+        /// an underscore when it seeds the value definitions. Normalising only the underscore silently broke every
+        /// hyphenated value: "Non-Compliant" never matched its own shortName, so a non-compliant item rolled up as
+        /// passed and the close-out rule could not see the shortfall it exists to catch.
+        /// </remarks>
+        public static bool AreSameEnumValue(string left, string right)
```
samatstariongroup, line 311: **[correctness, PLAUSIBLE]** Status classification uses hardcoded names, not enum refs

Status/compliance comparisons resolve to raw normalized text (case-insensitive, hyphen/space-folded) against hardcoded literal arrays instead of resolving to the actual EnumerationValueDefinition, so RDL customization or localization can silently desync classification.

**Failure scenario:** A user renames an EnumerationValueDefinition's Name (e.g. localizing 'Passed') via the RDL editor; the hardcoded ClosedPositiveStatuses/ConcludedStatuses literal-string lists no longer match, silently reclassifying passed items as open in every roll-up and rule with no error surfaced.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
```diff
@@ -0,0 +1,1444 @@
...
+
+            // the analysis check reads live parameter values, so a design change has to re-run it; ValueSet changes
+            // are what move a parameter, and they never touch the V&V item itself.
+            // Only a parameter some item actually covers is worth re-checking: re-running the check over the whole
+            // register on every value edit in the model stuttered the UI on models with a few hundred items
+            this.Disposables.Add(
+                this.CDPMessageBus.Listen<ObjectChangedEvent>(typeof(ParameterValueSet))
+                    .Merge(this.CDPMessageBus.Listen<ObjectChangedEvent>(typeof(ParameterOverrideValueSet)))
+                    .Where(x => x.ChangedThing.GetContainerOfType<Iteration>() == this.Thing)
+                    .Where(x => VandVCoverageQuery.IsCoveredParameter(this.Thing, x.ChangedThing.Container as ParameterOrOverrideBase))
```
samatstariongroup: **[efficiency, PLAUSIBLE]** Uncached O(n) coverage scan runs on every value-set change

The parameter-value-set change subscription is filtered only after calling `IsCoveredParameter`, which does an uncached O(relationships × categories) linear scan on every ParameterValueSet/ParameterOverrideValueSet change anywhere in the iteration, not just V&V-relevant ones.

**Failure scenario:** On a large model with thousands of relationships, every unrelated parameter edit anywhere in the model (not just V&V items) pays a full relationship-scan cost in the V&V browser's event handler, adding a constant per-edit tax that compounds during collaborative sessions with frequent value-set updates.

Requirements/Services/VandVItemCreator.cs
rdevoogtatstariongroup: Fixed: CanCreate now requires every category a write touches (coverage and procedure included) and the edit path gates on it too

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
rdevoogtatstariongroup: Fixed: creation and the reply/status actions are now gated on PermissionService.CanWrite

Requirements/Rdl/VandVRdlService.cs
rdevoogtatstariongroup: Fixed: Seed re-checks the library right before the write and seeds only what is still missing, so a concurrent set-up cannot duplicate reference data.

Requirements/ViewModels/VandV/VandVItemDialogViewModel.cs
rdevoogtatstariongroup: Fixed: the parameter is assigned before the element, so PopulateParameters sees it and keeps a stored ParameterOverride visible in the combo.

Requirements/Services/VandVItemCreator.cs
rdevoogtatstariongroup: Fixed: every category short-name now comes from one VandVCategory class, so a typo can no longer create a disconnected notion of coverage.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
rdevoogtatstariongroup: Fixed: the analysis check re-runs only when the changed value belongs to a parameter some V&V item actually covers.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
rdevoogtatstariongroup: Fixed: only the register's own link categories rebuild the coverage tree; an ordinary trace link no longer disposes every row.

Requirements/ViewModels/VandV/VandVMatrixViewModel.cs
rdevoogtatstariongroup: Fixed: the matrix subscriptions now filter on their own iteration, so an edit in another open model no longer rebuilds this panel.

Requirements/Services/VandVCoverageQuery.cs
rdevoogtatstariongroup: Fixed: a shortfall closed out without an accepted concession now counts as failed in the roll-up, matching what VnVItemCompletenessRule reports for the same item.

Requirements/Services/VandVItemCreator.cs
rdevoogtatstariongroup: Fixed: both copies are gone, replaced by a single VandVCoverageQuery.ResolveCategory all write paths call.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
rdevoogtatstariongroup: Fixed: ComputePermission now builds the set of annotation ClassKinds the participant may write and each create command gates on its own kind, so a change request is no longer enabled or disabled by the waiver permission.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
rdevoogtatstariongroup: Fixed: the item write is its own try block and the coverage/procedure writes another, so a failure after the item is committed now says the item was saved but its coverage and procedure were not, and tells the user to reopen and save it.

Requirements/ViewModels/VandV/VandVMatrixViewModel.cs
rdevoogtatstariongroup: Fixed: the fixed column names are reserved up front and a stage gate that collides with one gets its own numbered column, so the requirement identity can no longer be overwritten with coverage text

Requirements/Services/VandVItemCreator.cs
rdevoogtatstariongroup: Fixed: CanCreate now requires every parameter type in the manifest, because an attribute whose parameter type is absent is written as a silent no-op and the value the user typed would just disappear

Requirements/ViewModels/VandV/StageGateDialogViewModel.cs
rdevoogtatstariongroup: ixed: deduplication now runs on the derived short-name

Requirements/ViewModels/VandV/StageGateDialogViewModel.cs
rdevoogtatstariongroup: Fixed: the OK command now also listens to each row's Name, so blanking or duplicating names inline disables it immediately.

Requirements/ViewModels/VandV/Rows/VandVItemRowViewModel.cs
rdevoogtatstariongroup: Fixed: the drop event returns the write's task and Drop awaits it

Requirements/Rdl/VandVRdlManifest.cs
rdevoogtatstariongroup: Fixed: a VandVParameter class now declares all of them and every consumer references it, the same treatment the category short-names already had.

Requirements/Services/VandVCoverageQuery.cs
rdevoogtatstariongroup: Fixed: VandVStatus, VandVCompliance and VandVStepResult now drive both the seeded enumerations and every classification so the two cannot drift, with the note that renaming a seeded value definition in the RDL editor is not supported.

Requirements/ViewModels/VandV/VandVBrowserViewModel.cs
rdevoogtatstariongroup: Fixed: the covered parameter identifiers are cached and invalidated only when a V&V relationship changes, so an unrelated parameter edit costs a hash lookup instead of a full relationship scan.

samatstariongroup (approved):

PR #1478, opened 2026-08-21
Fixed #1475 error when deleting not the last item in an ordered list

### LINKED ISSUE #1475: [Bug] I get an error when deleting a middle enumeration value from an enumeration parameter
I get the following error when i try to delete an enumeration value from an enumeration parameter when it is not the last one .

Version 11.1.0-rc3

The CDP4 Services replied with code BadRequest: Bad Request: "exception:Could not resolve item EnumerationValueDefinition:6839a536-c3cd-41bd-817f-6fbf7aad5307"

Image

rdevoogtatstariongroup:
### Prerequisites

- [x] I have verified that there are no overlapping pull-requests open
- [x] I have verified that I am following the COMET-IME code style guidelines
- [x] I have provided test coverage for my change

### Description

Fix #1475 Error when deleting a middle enumeration value from an enumeration parameter

Deleting an `EnumerationValueDefinition` that was not the last one in the list failed with
`BadRequest: Could not resolve item EnumerationValueDefinition:<iid>`.

`UpdateTransaction` compared the clone's `OrderedItemList` against the dialog rows. A `Thing`
marked for deletion is not represented by a row, but it does remain in the list, so the
sequences differed and `OrderedItemList.Move` ran. `Move` rotates the values over a fixed set
of sort keys, which gave the deleted item a new key. The data-source cannot resolve an ordered
item that is deleted in the same transaction, hence the BadRequest. Deleting the last item
produced no `Move`, which is why only that case worked.

### Change

Added `DialogViewModelBase.UpdateOrderedItemList`, which keeps an item that is marked for
deletion on its current sort key and lays the rows out over the remaining positions. The 13
generated dialog view-models that reorder a composite `OrderedItemList` now call it instead of
each carrying its own copy of the loop, which removes 116 lines from the generated code.

The same defect was present when deleting a non-last `Component`, `PossibleState`, `Option`,
`ParametricConstraint`, `UnitFactor`, `QuantityKindFactor`, `RuleVerification`, `Section`,
`Page`, `Note`, `Book` and `Point`. Reordering without a deletion is unaffected.

### DIFF
--- a/CDP4Composition/CommonView/AutoGenDialogViewModel/DerivedUnitDialogViewModel.cs
+++ b/CDP4Composition/CommonView/AutoGenDialogViewModel/DerivedUnitDialogViewModel.cs
@@ -176,7 +176,20 @@ namespace CDP4CommonView
             var clone = this.Thing;
 
 
-            this.UpdateOrderedItemList(clone.UnitFactor, this.UnitFactor.Select(x => x.Thing));
+            if (!clone.UnitFactor.SortedItems.Values.SequenceEqual(this.UnitFactor.Select(x => x.Thing)))
+            {
+                var itemCount = this.UnitFactor.Count;
+                for (var i = 0; i < itemCount; i++)
+                {
+                    var item = this.UnitFactor[i].Thing;
+                    var currentIndex = clone.UnitFactor.IndexOf(item);
+
+                    if (currentIndex != i)
+                    {
+                        clone.UnitFactor.Move(currentIndex, i);
+                    }
+                }
+            }
             
         }
 
--- a/CDP4Composition/CommonView/AutoGenDialogViewModel/DerivedUnitDialogViewModel.cs
+++ b/CDP4Composition/CommonView/AutoGenDialogViewModel/DerivedUnitDialogViewModel.cs
@@ -176,20 +176,7 @@ namespace CDP4CommonView
             var clone = this.Thing;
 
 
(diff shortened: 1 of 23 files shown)


CDP4Composition/CommonView/AutoGenDialogViewModel/DerivedUnitDialogViewModel.cs
```diff
@@ -176,20 +176,7 @@ protected override void UpdateTransaction()
...
-                    var item = this.UnitFactor[i].Thing;
-                    var currentIndex = clone.UnitFactor.IndexOf(item);
-
-                    if (currentIndex != i)
-                    {
-                        clone.UnitFactor.Move(currentIndex, i);
-                    }
-                }
-            }
+            this.UpdateOrderedItemList(clone.UnitFactor, this.UnitFactor.Select(x => x.Thing));
```
samatstariongroup: have you updated the COMET codegenerator for this? or are these changes do localy?

samatstariongroup (changes requested):

CDP4Composition/CommonView/AutoGenDialogViewModel/DerivedUnitDialogViewModel.cs
rdevoogtatstariongroup: Oops my bad, totally missed this.

rdevoogtatstariongroup: Now incorporate the fix in another way outside of the generated files

PR #1485, opened 2026-08-27
Fix #1470 with additional function to set or add graph item to root items and added shortkey for edit and inspect

### LINKED ISSUE #1470: Traceability diagram: perf/reuse cleanup follow-ups from #1467 review
Follow-up cleanup/perf items found during review of #1467 (Configurable relationship traceability diagram, #1466). None of these are blocking — filed here to track separately rather than holding up the PR.

## Perf

- **Depth/MaxNodes edits rebuild graph on every keystroke** — `RelationshipTraceabilityViewModel.cs` (`DepthDown`/`DepthUp`/`MaxNodes` setters, ~line 295) call `ComputeGraph()` unconditionally on every set, and are bound in `RelationshipTraceability.xaml` with `UpdateSourceTrigger=PropertyChanged`. Typing "5000" into the MaxNodes SpinEdit fires 4 full synchronous graph rebuilds (5, 50, 500, 5000) — each doing a BFS traversal and rebuilding the `Nodes`/`Edges` `ReactiveList`s. Should debounce/throttle before recomputing.

- **Root query scans entire session cache on every recompute** — `QueryRootsByClassKindAndCategory` (`RelationshipTraceabilityViewModel.cs`, ~line 636) linearly scans the entire session `Assembler.Cache` (all `Thing`s across every open iteration/model), and runs on every `ComputeGraph()` call. Combined with the above, becomes the dominant cost of a single digit edit in a depth field on a large model with multiple iterations open.

- **Graph invalidation not scoped to the panel's iteration** — the `BinaryRelationship`/`MultiRelationship`/`Category` message-bus subscriptions that invalidate and recompute the graph (`RelationshipTraceabilityViewModel.cs`, ~line 855) are registered by `Type` only, not scoped to this panel's iteration/model. With two or more iterations open, editing a relationship/category in an unrelated iteration still triggers a full rebuild of this panel's graph.

## Reuse / duplication

- **Saved-config CRUD duplicates existing generic dialog infra** — `LoadSavedConfigurations`/`SaveConfiguration`/`DeleteConfiguration`/`ApplyConfiguration` (`RelationshipTraceabilityViewModel.cs`, ~line 930) hand-roll named-preset CRUD instead of reusing `SavedConfigurationDialogViewModel<T>`/`ManageConfigurationsDialogViewModel<T>`, which `RelationshipMatrixViewModel` already uses for the identical pattern (validated name/description entry, async save with busy/error state, dedicated manage/delete dialog). The hand-rolled version has no description field and no async/error handling.

- **SVG export duplicates level-color mapping** — `TraceabilityDiagramBehavior.BuildSvg()` (~line 197) hardcodes node fill colors as string literals that are exact RGB duplicates of the brushes already defined in `TraceabilityLevelToBrushConverter`. A future color-scheme change requires updating both places; missing one desyncs the exported SVG from the on-screen diagram.

- **ClassKind defaults duplicated from Relationship Matrix** — `DiagramEditorPluginSettings.DefaultClassKinds` (~line 45) is a verbatim copy of `RelationshipMatrixPluginSettings.DefaultClassKinds` (same doc comment even says so). The two lists will silently drift over time.

- **New list converters duplicate `GenericObjectListConverter<T>`** — `TraceabilityListConverter<T>` and its subclasses (`Helpers/TraceabilityListConverter.cs`, ~line 39) reimplement the same IList-to-`List<T>` cast logic already provided by `CDP4Composition.Converters.GenericObjectListConverter<T>`. Could reuse `GenericObjectListConverter<Category>`/`<ClassKind>` directly.

## Correctness (edge case, non-blocking)

- **Stale `Behavior` reference on DataContext reuse** — `TraceabilityDiagramBehavior.OnDataContextChanged` (~line 216) sets `viewModel.Behavior` on the new DataContext but never clears it on the previous view-model; `OnDetaching` only clears it on whatever DataContext is current at detach time. If a docked panel/tab is reused for a different traceability view instead of recreated, the old view-model keeps a live `Behavior` reference now driving a different diagram, so a later call from the stale view-model (e.g. `LayoutDirection` setter invoking `Behavior?.ApplyLayout()`) would apply layout/export to the wrong diagram.

_Filed from an automated review pass on PR #1467._

rdevoogtatstariongroup:
### Prerequisites

### Performance

- **Depth and MaxNodes rebuilt the graph on every keystroke.** The three spin editors now use `ValidateOnTextInput="False"` so typing "5000" commits once instead of rebuilding for 5, 50, 500 and 5000, and the setters skip the recompute when the value did not actually change.
- **The root query scanned the whole session cache.** `QueryRootsByClassKindAndCategory` now walks `Iteration.QueryContainedThingsDeep()` instead of every `Thing` of every open iteration and model.
- **Graph invalidation was not scoped to the panel's iteration.** The `BinaryRelationship` and `MultiRelationship` subscriptions now filter on the changed thing's containing iteration, so editing a relationship in another open iteration no longer rebuilds this panel.

### Reuse and duplication

- **Saved preset CRUD was hand rolled.** Saving and deleting now go through the shared `SavedConfigurationDialogViewModel<T>` and `ManageConfigurationsDialogViewModel<T>`, which the Relationship Matrix already uses, so presets get a description field and async save with busy and error state for free.
- **The SVG export duplicated the node colours.** `TraceabilityDiagramBehavior.BuildSvg` now reads `TraceabilityLevelToBrushConverter.GetHexColor(level)`, making the converter the single source of the level palette.
- **The ClassKind defaults were a verbatim copy of the Relationship Matrix list.** Both plugins now point at the new `CDP4Composition.RelationshipClassKinds.Default`.
- **The new list converters reimplemented existing logic.** `TraceabilityListConverter<T>` is deleted and the two subclasses derive from `CDP4Composition.Converters.GenericObjectListConverter<T>`.

### Correctness

- **A reused panel kept a stale `Behavior` reference.** `OnDataContextChanged` now clears `Behavior` on the previous view-model before injecting itself into the new one.

### Additionally added

- **Ctrl+E and Ctrl+I** on the diagram open the update and inspect dialog of the selected node or connector, matching the shortcuts the V&V browser already uses. The context menu labels advertise them.
- **Set as root** in the right click menu restarts the traversal from the selected node, which is how you check whether a node deeper down the tree also has other parents.
- **Add to roots** does the same but keeps the roots that are already shown, so a second branch expands next to the current one.
- **The maximum node warning now names the numbers**, showing the maximum that was hit and the range it may be raised within, and the spin editor bounds bind to those same two constants so they cannot drift apart.


CDP4Grapher.Tests/Helpers/TraceabilityLevelToBrushConverterTestFixture.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            var root = converter.Convert(0, typeof(Brush), null, CultureInfo.InvariantCulture);
+            var down = converter.Convert(2, typeof(Brush), null, CultureInfo.InvariantCulture);
+            var up = converter.Convert(-2, typeof(Brush), null, CultureInfo.InvariantCulture);
+
+            Assert.That(root, Is.Not.EqualTo(down));
+            Assert.That(down, Is.Not.EqualTo(up));
+
+            // a non-integer level still yields a usable brush rather than throwing on the diagram
+            Assert.That(converter.Convert(null, typeof(Brush), null, CultureInfo.InvariantCulture), Is.EqualTo(down));
```
joao4all: Pls assert multiple (also check the VerifyThatTheHexColorMatchesTheBrushOfTheSameLevel)

CDP4Grapher.Tests/ViewModels/RelationshipTraceabilityViewModelTestFixture.cs
```diff
@@ -791,21 +814,183 @@ public async Task VerifyThatAConfigurationCanBeSavedAppliedAndDeleted()
...
+
+            viewModel.SelectedNode = viewModel.Nodes.Single(x => x.Thing == this.spec2);
+            await viewModel.SetSelectedNodeAsRootCommand.Execute();
+
+            // the picked node replaces the roots, so the other path leading into it shows up above it
+            Assert.That(viewModel.RootThings, Is.EquivalentTo(new Thing[] { this.spec2 }));
+            Assert.That(viewModel.Nodes.Select(x => x.Thing), Is.EquivalentTo(new Thing[] { this.spec1, this.spec2, this.spec3 }));
+            Assert.That(viewModel.Nodes.Single(x => x.Thing == this.spec2).IsRoot, Is.True);
+
+            viewModel.Dispose();
```
joao4all, line 869: Not related to your PR itself, but previous implementation of this class is creating one viewmodel per test method. Do you think it is possible to create the viewmodel in the setup, and then dispose it in [TearDown] method? Just a suggestion, not blocking for merge the PR

CDP4Grapher/Settings/GrapherPluginSettings.cs
```diff
@@ -38,21 +34,5 @@ namespace CDP4Grapher.Settings
...
-            ClassKind.ElementUsage,
-            ClassKind.NestedElement,
-            ClassKind.Option,
-            ClassKind.Parameter,
-            ClassKind.ParametricConstraint,
-            ClassKind.RequirementsSpecification,
-            ClassKind.RequirementsGroup,
-            ClassKind.Requirement
-        };
     }
```
joao4all, line 57: Is this class still needed? I see it is empty, no more settings inside it.

joao4all (approved):
LGTM!!!

CDP4Grapher.Tests/Helpers/TraceabilityLevelToBrushConverterTestFixture.cs
rdevoogtatstariongroup: Fixed it

CDP4Grapher/Settings/GrapherPluginSettings.cs
rdevoogtatstariongroup: We need to keep it, because The CDP4Grapher looks for it

CDP4Grapher.Tests/ViewModels/RelationshipTraceabilityViewModelTestFixture.cs
rdevoogtatstariongroup: left as-is because the view-model constructor computes the graph and therefore has to be created after each test's data.

PR #1488, opened 2026-09-01
#1487 Fixed the ReqIF Exporter to not have a specobjectType for each requirement.

### LINKED ISSUE #1487: [Bug] When I export a reqIF it makes for every requirement a specObjectType instead of one.
When I export a reqIF it makes for every requirement a specObjectType instead of the logical amount of probably one or few.

rdevoogtatstariongroup:
### Prerequisites

### Description
#1487 Fixed the ReqIF Exporter to not have a specobjectType for each requirement. It makes a specObjectType per different combination of rules and simple parameter values.

### DIFF
--- a/Requirements.Tests/ReqIF/ParameterTypeMappingDialogTestFixture.cs
+++ b/Requirements.Tests/ReqIF/ParameterTypeMappingDialogTestFixture.cs
@@ -1,6 +1,6 @@
 ﻿// --------------------------------------------------------------------------------------------------------------------
 // <copyright file="ParameterTypeMappingDialogTestFixture.cs" company="Starion Group S.A.">
-//    Copyright (c) 2015-2024 Starion Group S.A.
+//    Copyright (c) 2015-2026 Starion Group S.A.
 //
 //    Author: Sam Gerené, Alex Vorobiev, Alexander van Delft, Nathanael Smiechowski, Antoine Théate, Omar Elebiary
 //
@@ -29,6 +29,7 @@ namespace CDP4Requirements.Tests.ReqIF
     using System.Collections.Generic;
     using System.Linq;
     using System.Reactive.Linq;
+    using System.Text.RegularExpressions;
     using System.Threading.Tasks;
 
     using CDP4Common.CommonData;
@@ -181,6 +182,20 @@ namespace CDP4Requirements.Tests.ReqIF
         }
 
         [Test]
+        public void VerifyThatEnumValueRowsHaveNoCreateCommands()
+        {
+            var enumRow = this.dialog.MappingRows.First(x => x.Identifiable == this.enumDatadef);
+
+            // the enumeration datatype row itself can create an Enumeration Parameter Type
+            this.dialog.SelectedRow = enumRow;
+            Assert.AreEqual(1, this.dialog.CreateParameterTypeCommands.Count);
+
+            // but its enum-value child rows (the literals) cannot create a parameter type
+            this.dialog.SelectedRow = enumRow.EnumValue.First();
+            Assert.AreEqual(0, this.dialog.CreateParameterTypeCommands.Count);
+        }
+
+        [Test]
         public void VerifyThatCreateGenericParameterTypeWorks()
         {
             this.thingDialogNavigationService.Setup(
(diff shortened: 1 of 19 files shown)


Requirements.Tests/ReqIF/ParameterTypeMappingDialogTestFixture.cs
```diff
@@ -180,6 +181,20 @@ public void VerifyThatMenuIsPopulated()
...
         }
 
+        [Test]
+        public void VerifyThatEnumValueRowsHaveNoCreateCommands()
+        {
+            var enumRow = this.dialog.MappingRows.First(x => x.Identifiable == this.enumDatadef);
+
+            // the enumeration datatype row itself can create an Enumeration Parameter Type
+            this.dialog.SelectedRow = enumRow;
+            Assert.AreEqual(1, this.dialog.CreateParameterTypeCommands.Count);
```
joao4all: Suggestion: using Assert.That(list, Has.Count.EqualTo(...); is more idiomatic and follows the Constraint-Based Model for nunit tests.

This comment applies to every place where Assert.IsFalse/True/AreEqual/IsNotNull/etc... is used in this PR

Requirements/ReqIFDal/ReqIFBuilder.cs
```diff
@@ -685,5 +723,45 @@ private void BuildToolExtension()
         {
             //TODO
         }
+
+        /// <summary>
+        /// Compares two sets of <see cref="ParameterizedCategoryRule"/>s for equality, ignoring order, so that
+        /// requirements can be grouped by their applied rule-set.
+        /// </summary>
+        private sealed class RuleSetEqualityComparer : IEqualityComparer<ParameterizedCategoryRule[]>
```
joao4all: Pls move this to its own file

Requirements/ReqIFDal/ThingFactory.cs
```diff
@@ -680,5 +689,31 @@ private string GetAttributeValue(AttributeValue value)
...
         }
+
+        /// <summary>
+        /// Converts an XHTML (rich-text) attribute value to readable plain text. Line-breaking elements become new
+        /// lines, the remaining tags are removed and the XML/HTML entities are decoded. This keeps the textual content
+        /// of DOORS/Capella (and other) XHTML attributes instead of importing the raw markup.
+        /// </summary>
+        /// <param name="xhtml">The XHTML markup</param>
+        /// <returns>The plain-text representation</returns>
+        internal static string ConvertXhtmlToText(string xhtml)
```
joao4all: 3 comments about this method
- Why specficially internal and not private?
- Regex is usually slow, and for this case, where I imagine this method will be used many times, maybe you should consider a different approach, like XmlReader (then HtmlDecode might be removed).
- If you decide to keep regex, please use it as a source generated fields. Also check for edge cases in this method tests, as we may encounter <style> or <script> tags (or more edgy things) in the xhtml, I suppose

Requirements/ViewModels/ReqIF/ReqIfExportDialogViewModel.cs
```diff
@@ -149,11 +156,19 @@ public ReqIfExportDialogViewModel(IEnumerable<ISession> sessions, IEnumerable<It
                 this.Iterations.Add(new ReqIfExportIterationRowViewModel(iteration));
             }
 
+            this.WhenAnyValue(vm => vm.SelectedIteration).Subscribe(_ => this.PopulateRequirementsSpecifications());
+
+            this.WhenAnyValue(vm => vm.IncludeDeprecated).Subscribe(_ => this.PopulateRequirementsSpecifications());
+
```
joao4all: The subscriptions in this class should be added to the disposables list

Requirements/ViewModels/ReqIF/ReqIfExportDialogViewModel.cs
```diff
@@ -328,11 +382,108 @@ private async Task<ReqIF> BuildReqIf()
...
+            }
+
+            try
+            {
+                this.IsBusy = true;
+                this.LoadingMessage = "Building the type preview...";
+
+                var session = this.Sessions.Single(x => x.Assembler.Cache == this.SelectedIteration.Iteration.Cache);
+
+                var reqif = await Task.Run(() => new ReqIFBuilder().BuildReqIF(session, this.SelectedIteration.Iteration, this.IncludeDeprecated, selectedRequirementsSpecifications, this.SelectedFormat));
```
joao4all, line 420: Why the task.run? Wouldn't be better to run it in the same thread?

Requirements/ViewModels/ReqIF/ReqIfImportDialogViewModel.cs
```diff
@@ -302,7 +302,7 @@ private void ExecuteCancel()
         /// </summary>
         private void ExecuteBrowse()
         {
-            var result = this.fileDialogService.GetOpenFileDialog(true, true, false, "ReqIF files (*.reqif, *.xml)|*.reqif; *.xml|All Files (*.*)|*.*", ".reqif", this.Path, 1);
+            var result = this.fileDialogService.GetOpenFileDialog(true, true, false, "ReqIF files (*.reqif, *.reqifz, *.zip)|*.reqif; *.reqifz; *.zip|All Files (*.*)|*.*", ".reqif", this.Path, 1);
```
joao4all: Just a question to make sure: previously the code was using .xml files too. Now it doesn't. Is this intended?

joao4all (changes requested):

Requirements.Tests/ReqIF/ParameterTypeMappingDialogTestFixture.cs
rdevoogtatstariongroup: Fixed it

Requirements/ReqIFDal/ReqIFBuilder.cs
rdevoogtatstariongroup: moved it to its own file

Requirements/ViewModels/ReqIF/ReqIfExportDialogViewModel.cs
rdevoogtatstariongroup: added them

Requirements/ViewModels/ReqIF/ReqIfExportDialogViewModel.cs
rdevoogtatstariongroup: This keeps the UI responsive and able to cancel

Requirements/ViewModels/ReqIF/ReqIfImportDialogViewModel.cs
rdevoogtatstariongroup: Restored *.xml

Requirements/ReqIFDal/ThingFactory.cs
rdevoogtatstariongroup: Made ConvertXhtmlToText private and rewrote the test to exercise it through the public ComputeRequirementThings path (an XHTML requirement value asserted on the resulting Requirement.Definition); on the other two points I kept regex rather than XmlReader because DOORS/Capella markup isn't guaranteed to be well-formed XML (an XmlReader would throw), but hoisted the patterns into static readonly RegexOptions.Compiled fields — [GeneratedRegex] source generators aren't available on net48 — and it now drops <script>/<style> with their content, covered by the edge-case assertions.

joao4all (approved):

PR #1489, opened 2026-09-01
Fix #1480 and #1481 full error text in details and ruleVerificationDialog always active on creation

### LINKED ISSUE #1480: Rule Verification Lists - when adding new rules to list, set Active checkbox as checked by default (so as active rule)
Rule Verification Lists - when adding new rules to list, set Active checkbox as checked by default (so as active rule)

now a new rule is not set as active, leading to excluding the newly added rule from execution, only after some time realising that active is not set properly.

noticed when verifying #1275

### LINKED ISSUE #1481: Rule Verification Lists - violation details in field "Name" more explicit now, but still hard to read completely as long string of text
Rule Verification Lists - violation details in field "Name" more explicit now, but still hard to read completely as long string of text

related to #1275 already more clear now (with indication of e.g. parameter short names). 

with multiple violations however, the text can become quite long, so it is still pretty difficult to read this within the frame of the Name-field where this is shown. Also expanding the browser setting this field as wide as possible is not always sufficient to completely read it. 

so possible improvement here would still be welcome (e.g. adding the violating text to the details box in the bottom of the browser instead of only the Type (RuleViolation) that is given now. this seems a logical place to add this text as well.

rdevoogtatstariongroup:
### Prerequisites

### Description
Fix #1480 and #1481 full error text in details and ruleVerificationDialog always active on creation

### DIFF
--- a/EngineeringModel/ViewModels/Dialogs/UserRuleVerificationDialogViewModel.cs
+++ b/EngineeringModel/ViewModels/Dialogs/UserRuleVerificationDialogViewModel.cs
@@ -120,6 +120,8 @@ namespace CDP4EngineeringModel.ViewModels
 
             if (this.dialogKind == ThingDialogKind.Create)
             {
+                this.IsActive = true;
+
                 var container = (RuleVerificationList)this.Container;
                 this.Owner = string.Format("{0} [{1}]", container.Owner.Name, container.Owner.ShortName);
             }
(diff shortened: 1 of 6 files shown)


EngineeringModel/ViewModels/Dialogs/UserRuleVerificationDialogViewModel.cs
```diff
@@ -120,6 +120,8 @@ protected override void UpdateProperties()
 
             if (this.dialogKind == ThingDialogKind.Create)
             {
+                this.IsActive = true;
```
joao4all, line 123: Question: does this need to be set to false at any point?

joao4all (approved):
LGTM

EngineeringModel/ViewModels/Dialogs/UserRuleVerificationDialogViewModel.cs
rdevoogtatstariongroup: TBH, I think you probably never do it, but as it is a property on the rule we leave the option to set it in. This also gives you the option to temporarily verify less rules without changing the list.

PR #1492, opened 2026-09-07
Feat/gh1486 create and delete rel in traceability

### LINKED ISSUE #1486: [Feature] As a user I want to be able to Create and delete relationships from the traceability diagram
As a user I want to be able to Create and delete relationships from the traceability diagram.

rdevoogtatstariongroup:
### Prerequisites

### #1486 Create and delete relationships from the traceability diagram

Makes the Relationship Traceability panel (CDP4Grapher) editable and more configurable.

### Create / delete relationships (DOORS-style)
- Right-click a node → **Start link from here** (the source node is outlined and a banner is shown; Esc cancels).
- Right-click a target node → **Create link from start** / **Create link to start** → pick the link kind.
- Link kinds are driven **only** by the `BinaryRelationshipRule`s whose source/target categories match the two elements; if none apply the submenu shows a disabled *"No applicable relationship rule"*.
- Right-click a relationship → **Delete relationship** (with confirmation).

### Configurable block content
- New **Block content** config group: toggle Class kind / Short name / Name / Definition per node, plus a definition max-length cap (truncated with an ellipsis). Persisted in saved presets.

### Details tab
- New **Details** tab in the config pane showing copy-pasteable properties of the selected node/relationship (kind, name, short name, owner, categories, endpoints, definition).

### Misc
- Standard DevExpress glyphs added to the context menu (edit, inspect, delete, export, linking).

### DIFF
--- a/CDP4Grapher/ViewModels/RelationshipTraceabilityViewModel.cs
+++ b/CDP4Grapher/ViewModels/RelationshipTraceabilityViewModel.cs
@@ -53,6 +53,7 @@ namespace CDP4Grapher.ViewModels
 
     using CDP4Dal;
     using CDP4Dal.Events;
+    using CDP4Dal.Operations;
 
     using CDP4Grapher.Behaviors;
     using CDP4Grapher.Helpers;
@@ -189,6 +190,41 @@ namespace CDP4Grapher.ViewModels
         /// Backing field for <see cref="SelectedConfiguration"/>
         /// </summary>
         private TraceabilityConfiguration selectedConfiguration;
+
+        /// <summary>
+        /// Backing field for <see cref="ShowClassKind"/>
+        /// </summary>
+        private bool showClassKind = true;
+
+        /// <summary>
+        /// Backing field for <see cref="ShowShortName"/>
+        /// </summary>
+        private bool showShortName = true;
+
+        /// <summary>
+        /// Backing field for <see cref="ShowName"/>
+        /// </summary>
+        private bool showName = true;
+
+        /// <summary>
+        /// Backing field for <see cref="ShowDefinition"/>
+        /// </summary>
+        private bool showDefinition;
+
+        /// <summary>
+        /// Backing field for <see cref="DefinitionMaxLength"/>
+        /// </summary>
+        private int definitionMaxLength = 100;
+
+        /// <summary>
+        /// Backing field for <see cref="LinkSourceThing"/>
+        /// </summary>
+        private Thing linkSourceThing;
+
+        /// <summary>
+        /// Backing field for <see cref="SelectedThingDetails"/>
+        /// </summary>
+        private string selectedThingDetails;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RelationshipTraceabilityViewModel"/> class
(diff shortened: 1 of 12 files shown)


CDP4Grapher/ViewModels/RelationshipTraceabilityViewModel.cs
```diff
@@ -525,6 +561,113 @@ public TraceabilityConfiguration SelectedConfiguration
...
+        /// <summary>
+        /// Gets or sets a value indicating whether the node boxes show the class kind line
+        /// </summary>
+        public bool ShowClassKind
+        {
+            get => this.showClassKind;
+            set
+            {
+                this.RaiseAndSetIfChanged(ref this.showClassKind, value);
+                this.ComputeGraph();
```
joao4all: Why not subscribing (this.whenanyvalue....) to the changes in ShowClassKind (reactive property), instead of calling ComputeGraph in its setter?

When subscribing, we can take advantage of using the reactiveUI property, otherwise it could be just a regular property. I would go for reactiveUI subscription, but is up to you, np

This comment applies to all new properties for this class in this PR

CDP4Grapher/ViewModels/RelationshipTraceabilityViewModel.cs
```diff
@@ -949,6 +1129,198 @@ private void InitializeTraceabilityCommands()
                         this.Behavior?.Export(format);
                     }
                 });
+
+            this.StartLinkCommand = ReactiveCommandCreator.Create(
```
joao4all, line 1108: Question: don't we need to use "disposewith" at the end of this declaration? Or put it inside the disposables list?

CDP4Grapher/ViewModels/RelationshipTraceabilityViewModel.cs
joao4all, line 1: This class is almost 1600 lines of code. It would be nice if you could do some refactor to shrink a bit its size. Maybe it's better to create a ticket for this, or doing in this PR, idk, up to you :)

joao4all (approved):

CDP4Grapher/ViewModels/RelationshipTraceabilityViewModel.cs
rdevoogtatstariongroup: Partially  I pulled the link-creation logic and the details-text builder into dedicated Helpers classes (RelationshipLinkHelper, TraceabilityDetailsBuilder) and moved LinkCreationOption alongside them. but all the save and apply is still in the viewmodel.

CDP4Grapher/ViewModels/RelationshipTraceabilityViewModel.cs
rdevoogtatstariongroup: Good call, I fixed it with one whenanyvalue

CDP4Grapher/ViewModels/RelationshipTraceabilityViewModel.cs
rdevoogtatstariongroup: the subscription is wrapped in this.Disposables.Add(...)

